PR #60, opened 2023-01-17
Nmb 19 add vetoing

DorianSternVukotic:
Adds ability to veto and revert veto.

This is a bit unoptimized, specifically getLegislation in noderpc edits, because to get vetos, blockchain stores all adresses that vetoed it, we need only the number of vetoes and to see if user did veto.
But to get this, since it doesnt track numbers of vetos directly, we need to use entries (api.query.liberlandLegislation.vetos.entries)

which cannot be combined with .multi, so there are many api queries. Left a blockchain task to optimize this.

Also, doesnt yet have a loading wheel, that will come as a separate task.

### DIFF
--- a/src/components/Legislation/LegislationView/index.js
+++ b/src/components/Legislation/LegislationView/index.js
@@ -2,14 +2,17 @@ import React, { useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useParams } from 'react-router-dom';
 import { legislationActions } from '../../../redux/actions';
-import { legislationSelectors } from '../../../redux/selectors';
+import { blockchainSelectors, legislationSelectors } from '../../../redux/selectors';
 import Card from '../../Card';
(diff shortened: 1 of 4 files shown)


src/components/Legislation/LegislationView/index.js
```diff
@@ -19,8 +22,33 @@ const LegislationView = () => {
...
+      <div className={styles.legislationInfoContainer}>
+        <div className={styles.legislationContent}>
+          {l.content}
+        </div>
+        <div className={styles.vetoContent}>
+          <div className={styles.vetoInfo}>
+            <div>
+              <b>{l?.vetos?.length}</b>
+              {' '}
+              / xyz
```
kacperzuk-neti, line 38: Some leftover placeholder, not sure if it's left as is "for now" or forgotten.

kacperzuk-neti (approved):

kacperzuk-neti:
Efficient veto count on blockchain: https://github.com/liberland/liberland_substrate/pull/222/files

PR #61, opened 2023-01-22
Error and loading handling for blockchain actions

DorianSternVukotic:


src/api/nodeRpcCall.js
```diff
@@ -90,45 +95,47 @@ const sendTransferLLM = async (payload, callback) => {
...
-      if (status.isInBlock) {
-        // eslint-disable-next-line no-console
-        console.log(`InBlock at block hash #${status.asInBlock.toString()}`);
-        callback(null, 'done');
-      }
-    }).catch((error) => {
+  const { values: { amount }, isUserHaveStake, walletAddress } = payload;
+  const api = await getApi();
+  const transferExtrinsic = isUserHaveStake
+    ? await api.tx.staking.bondExtra(`${amount}000000000000`)
```
kacperzuk-neti, line 119: How about using `dollarsToGrains` from `walletHelpers.js`?
This looks risky - what if amount is "15.0"?

kacperzuk-neti: Additionally, although this is subjective, so treat this as optional - converting to grains should probably be done ASAP at the input, not deep down in nodeRpcCall.js.

PR #93, opened 2023-07-31
BLOCKCHAIN-101 - Make a persistent subscription to best block number

kacperzuk-neti:


DorianSternVukotic:
Why

kacperzuk-neti:
1. In the bridge, after transfer is approved we need to wait until current block number reaches the funds unlock block number. Right now current block number is not being updated automatically, so unless the user refreshes the page it will be stuck on "Processing (~1h)` state.
2. In the identity office panel we use the current time + current block number to calculate candidate citizen's birthday. If the current block number is out of date this will be inaccurate.

This PR makes sure that the current block number in Redux Store is always up-to-date.

PR #113, opened 2023-08-30
BLOCKCHAIN-131 - FE - list all current congress motions

MichalLustykNeti:


src/components/Congress/Motions/index.js
```diff
@@ -0,0 +1,95 @@
...
+  const sender = useSelector(blockchainSelectors.userWalletAddressSelector);
+
+  useEffect(() => {
+    dispatch(congressActions.getMotions.call());
+  }, [dispatch]);
+
+  return (
+    <div>
+      {/* eslint-disable-next-line react/prop-types */}
+      {motions.map(({ proposal, proposalOf, voting }) => {
```
kacperzuk-neti, line 26: Extract it to a separate component (can be in the same file), so that we can have a more readable:
```
{
  motions.map(({ proposal, proposalOf, voting }) =>
    <Motion key={FIXME} proposal={proposal} proposalOf={proposalOf} voting={voting} />
  )
}
```

src/components/Congress/Motions/index.js
```diff
@@ -0,0 +1,95 @@
...
+                    primary
+                    onClick={() => dispatch(
+                      congressActions.voteAtMotions.call({
+                        readableProposal,
+                        index: readableVoting.index,
+                        vote: true,
+                      }),
+                    )}
+                  >
+                    Vote ayes
```
kacperzuk-neti, line 72:
```suggestion
                    Vote aye
```

src/components/Congress/Motions/index.js
```diff
@@ -0,0 +1,95 @@
...
+                    secondary
+                    onClick={() => dispatch(
+                      congressActions.voteAtMotions.call({
+                        readableProposal,
+                        index: readableVoting.index,
+                        vote: false,
+                      }),
+                    )}
+                  >
+                    Vote nays
```
kacperzuk-neti, line 85:
```suggestion
                    Vote nay
```

src/redux/sagas/congress.js
```diff
@@ -30,6 +35,36 @@ function* getCongressCandidatesWorker() {
...
+    readableProposal,
+    index,
+    vote,
+  );
+  if (errorData.isError) {
+    yield put(
+      blockchainActions.setErrorExistsAndUnacknowledgedByUser.success(true),
+    );
+    yield put(blockchainActions.setError.success(errorData));
+    yield put(congressActions.applyForCongress.failure(errorData));
```
kacperzuk-neti, line 61:
```suggestion
    yield put(congressActions.voteAtMotions.failure(errorData));
```

kacperzuk-neti (changes requested):
Address comments + squash

kacperzuk-neti:
Messed up squash/rebase, please fix.

src/api/nodeRpcCall.js
```diff
@@ -1305,6 +1305,25 @@ const getBlockEvents = async (blockHash) => {
   }
 }
 
+const getMotions = async () => {
+  const api = await getApi();
+  const proposals = await api.query.council.proposals()
+  
+  return await Promise.all(
+    proposals.map(async proposal => {
+      const [proposalOf, voting] = await api.queryMulti([
```
kacperzuk-neti, line 1314: So we're fetching N proposals and then we do another call per proposal, effectively doing N+1 calls. This is known as "N+1 query problem".

While this is OK here , as I don't expect many proposals, in the future in other cases consider one of these 2 solutions:
1. Fetch all proposals + all proposalOf (using .entries()) + all voting (using .entries()) and join them client side, which would make only 1 call to node, or
2. Fetch all proposals and then fetch all proposalOf and voting for these proposals using `.multi()`. As .multi returns result sorted the same way as incoming list, it's trivial to join them. That would only do 3 calls to node.

src/components/Congress/Motions/index.js
```diff
@@ -0,0 +1,126 @@
...
+                  readableProposal: proposal,
+                  index: voting.index,
+                  vote: true,
+                }),
+              )}
+            >
+              Vote aye
+            </Button>
+          )}
+          {!voting.nays.includes(sender) && (
```
kacperzuk-neti, line 90: Hmm... does the blockchain allow us to vote nay if we've already voted aye (and vice versa)? Does it change the vote, fail or is ignored?

kacperzuk-neti (approved):

src/components/Congress/Motions/index.js
MichalLustykNeti: We can not vote the same (it is handled in code by throwing error). But when you are trying to vote opposite to your already existing vote. So when we voted "nay" we can not vote "nay" but we can vote "aye"

DorianSternVukotic (approved):

PR #116, opened 2023-09-05
BLOCKCHAIN-133 - congress - form for proposing motion for new international treaty legislation

kacperzuk-neti:


src/components/Congress/ProposeLegislationButton/index.js
```diff
@@ -0,0 +1,16 @@
+import React, { useState } from 'react';
+import Button from '../../Button/Button';
+import CongressProposeLegislationModal from '../../Modals/CongressProposeLegislationModal';
+
+export default function ProposeLegislationButton() {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const handleModalOpen = () => setIsModalOpen(!isModalOpen);
```
kacperzuk-neti, line 7: FIXME `return null` if user is not a congressman

src/api/nodeRpcCall.js
```diff
@@ -1574,6 +1574,37 @@ const stakingBondExtra = async (value, walletAddress, callback) => {
   });
 }
 
+const getCongressMembers = async () => {
+  const api = await getApi();
+  const members = await api.query.council.members();
+  return members.map(m => m.toString());
```
MichalLustykNeti, line 1580: i believe toHoman will do the same

kacperzuk-neti: There are no guarantees on `.toHuman()`. It's output is lossy (strips most enums etc.) and can change with update or with different Locale. There're also typing issues: toHuman returns `AnyJson`, while this will return `Array<String>`, which will be PITA if we ever decide to start using TypeScript.
If you are lazy, you can use `.toJSON()`. While it still returns `AnyJson`, at least it's defined how it will serialize each type.
Ideally, `.toHuman()` should be only used for debugging and `.toJSON()` only for serialization.

See also this comment by one of devs: https://tinyurl.com/5743yvm2

src/api/nodeRpcCall.js
```diff
@@ -1583,6 +1583,37 @@ const stakingBondExtra = async (value, walletAddress, callback) => {
...
+}
+
+const congressProposeLegislation = async (tier, index, legislationContent, walletAddress, callback) => {
+  const api = await getApi();
+  const injector = await web3FromAddress(walletAddress);
+  const congressmen = await api.query.council.members();
+  const threshold = parseInt(congressmen.length/2) + 1;
```
DorianSternVukotic, line 1596: in the future, they should be able to plan the threshold in advance, as councilproposals can be different in what referendum they spawn. But will add it as a new task

DorianSternVukotic (approved):

PR #120, opened 2023-09-06
 BLOCKCHAIN-157 - congress - treasury tab 

kacperzuk-neti:


src/api/nodeRpcCall.js
```diff
@@ -1776,6 +1776,41 @@ const congressRepealLegislation = async (tier, index, walletAddress, callback) =
...
+}
+
+const getTreasurySpendPeriod = async () => {
+  const api = await getApi();
+  return api.consts.treasury.spendPeriod;
+}
+
+const getTreasuryBudget = async () => {
+  const api = await getApi();
+  const account = "5EYCAe5ijiYfyeZ2JJCGq56LmPyNRAKzpG4QkoQkkQNB5e6Z";
```
MichalLustykNeti, line 1791: can we use in here addres from https://github.com/liberland/liberland_frontend/pull/125 ?

kacperzuk-neti: 1. #125 is not merged yet
2. it would need to be refactored significantly to be reusable in different places. Something that would allow me to `dispatch(officesActions.getGovFinances.call())` and then `useSelector(officesSelectors.govFinances)` or smth.

So I think it's out of the scope for this PR.

DorianSternVukotic: buuuut it should be not hardcoded or at least stored in a variable, when we deploy to prod

DorianSternVukotic (approved):

PR #122, opened 2023-09-11
added Council addres

MichalLustykNeti:


kacperzuk-neti (approved):

DorianSternVukotic:
Is this different for mainnet/testnet? How do i verify this is the correct address ?

kacperzuk-neti:
> Is this different for mainnet/testnet? How do i verify this is the correct address ?

@DorianSternVukotic  It's the same for both. The address is derived from PalletID configured in runtime: https://github.com/liberland/liberland_substrate/pull/314/files#diff-67684005af418e25ff88c2ae5b520f0c040371f1d817e03a3652e76b9485224aR1613 , which is `councilc`. You can do the derivation manually on https://www.shawntabrizi.com/substrate-js-utilities/ , it's the `"Module ID" to Address` form - see screenshot.

image

I believe @MichalLustykNeti is working on small refactor of this module, so that:
* pallet id is fetched from the api metadata
* address is derived automatically from the pallet id

that way the only thing we'll have to hardcode is pallet name.

DorianSternVukotic (approved):

PR #124, opened 2023-09-12
BLOCKCHAIN-134 - FE - form for congressman to create new motion - spending from LLD treasury (pallet-treasury)

MichalLustykNeti:


src/components/Modals/TreasurySpendingMotionModal.js
```diff
@@ -0,0 +1,103 @@
...
+  const dispatch = useDispatch();
+
+  const {
+    handleSubmit,
+    register,
+    formState: { errors },
+  } = useForm({
+    mode: 'all',
+    defaultValues: {
+      spending: 'congressSendLlm',
```
kacperzuk-neti, line 25: Is this used anywhere?

src/components/Modals/TreasurySpendingMotionModal.js
```diff
@@ -0,0 +1,103 @@
...
+    },
+  });
+
+  const onSubmit = ({ transferToAddress, transferAmount }) => {
+    dispatch(
+      congressActions.congressSendTreasuryLld.call({
+        transferToAddress,
+        transferAmount,
```
kacperzuk-neti, line 33: This is wrong.

`transferAmount` is in LLD - so it's a string like `100.123`.
Sagas pass this directly to nodeRpcCall and nodeRpcCall passes it directly to polkadot-js-api, which directly passes it to blockchain.
So this will transfer `0.0000000001` LLD instead of the requested `100.123`.
You need `dollarsToGrains()` call here.

kacperzuk-neti: Also I see that we've missed the same error in previous reviews. Can you fix this in `SpendingMotionModal.js` as well? Can be in this PR.

src/components/Modals/TreasurySpendingMotionModal.js
```diff
@@ -0,0 +1,103 @@
...
+      )}
+      <span className={styles.title}>Amount</span>
+      <TextInput
+        register={register}
+        name="transferAmount"
+        required
+        errorTitle="Amount not valid"
+        validate={(v) => {
+          try {
+            if (valueToBN(v).lt(0)) {
```
kacperzuk-neti, line 68: Why `valueToBN` instead of standard parseFloat?
Also I see that we've missed the same thing in previous reviews. Can you fix this in `SpendingMotionModal.js` as well? Can be in this PR.

src/api/nodeRpcCall.js
```diff
@@ -1877,6 +1877,32 @@ const congressProposeLegislationReferendum = async (
...
+const congressSendTreasuryLld = async (transferToAddress, transferAmount, walletAddress, callback) => {
+  const api = await getApi();
+  const injector = await web3FromAddress(walletAddress);
+
+  const threshold = await congressMajorityThreshold();
+  const proposal = api.tx.treasury.spend(transferAmount, transferToAddress);
+  const length = proposal.length;
+  
+  const extrinsic = api.tx.council.propose(threshold, proposal, length);
+  extrinsic.signAndSend(walletAddress, { signer: injector.signer }, ({ status, events, dispatchError }) => {
```
kacperzuk-neti, line 1889: As we'll need updates to this PR either way, can you update this to new convention? So:
* `submitExtrinsic` here in nodeRpcCall
* worker in sagas that uses `call` instead of `cps` and doesn't do explicit error checking
* `blockchainWatcher` for watcher in sagas.

kacperzuk-neti (changes requested):

kacperzuk-neti (approved):

DorianSternVukotic (approved):

PR #151, opened 2023-10-09
BLOCKCHAIN-223 Bridge FE - balance is not updated after deposit

MichalLustykNeti:


src/redux/sagas/bridge.js
```diff
@@ -52,21 +52,27 @@ function* withdrawWorker(action) {
...
-        event.method == "OutgoingReceipt"
+      const receiptEventIndex = blockEvents.findIndex(
+        ({ phase, event }) => phase.isApplyExtrinsic
+        && phase.asApplyExtrinsic.eq(extrinsicIndex)
+        && event.method === 'OutgoingReceipt',
       );
       const { amount, ethRecipient } = blockEvents[receiptEventIndex].event.data;
-      let receipt_id = subToEthReceiptId(blockHash, receiptEventIndex, amount, ethRecipient);
+      const receipt_id = subToEthReceiptId(blockHash, receiptEventIndex, amount, ethRecipient);
+      yield put(walletActions.getWallet.call());
```
kacperzuk-neti, line 74: `put` this after the `bridgeActions.deposit` - otherwise the loader won't work as it should. (as the `deposit.success` will hide it before `getWallet` finishes).

kacperzuk-neti (changes requested):

kacperzuk-neti (approved):

src/redux/sagas/bridge.js
```diff
@@ -75,8 +80,10 @@ function* depositWorker(action) {
         blockHash,
         date: Date.now(),
       }));
+      yield put(walletActions.getWallet.call());
```
DorianSternVukotic, line 83: So, this is the only line required to satisfy the tickets, the others are just refactors xD

PR #152, opened 2023-10-09
BLOCKCHAIN-211 + BLOCKCHAIN-232 - Ugly, but user-readable motions/proposal/referenda + referenda in enactment period

kacperzuk-neti:


MichalLustykNeti (approved):

DorianSternVukotic:
I will merge this, but its possible that the current implementation makes using the application very slow, and we should watch out if real world referendums will cause this

PR #156, opened 2023-10-27
BLOCKCHAIN-240 + BLOCKCHAIN-246 + BLOCKCHAIN-247 - optimize company data format + small UX improvements + bugfixes

kacperzuk-neti:
Using sample data:

Original JSON encoding: 3279 bytes
JSON encoding + deflate compressed: 827 bytes

SCALE encoded: 1073 bytes
This PR - SCALE encoded + deflate compressed: 459 bytes

So for sample data we've reduced data size by 86%.

### DIFF
--- a/src/api/nodeRpcCall.js
+++ b/src/api/nodeRpcCall.js
@@ -409,9 +409,9 @@ const getCompanyRegistration = async (entity_id) => {
   }
 }
 
-const registerCompany = async ({company_id, hash, walletAddress}, callback) => {
+const registerCompany = async ({entity_id, hash, walletAddress}, callback) => {
   const api = await getApi();
-  const registerCall = api.tx.companyRegistry.registerEntity(0, company_id, hash);
+  const registerCall = api.tx.companyRegistry.registerEntity(0, entity_id, hash);
   const proxied = api.tx.companyRegistryOffice.execute(registerCall);
   const injector = await web3FromSource('polkadot-js');
   proxied.signAndSend(walletAddress, { signer: injector.signer }, ({ status, events, dispatchError }) => {
--- a/src/api/nodeRpcCall.js
+++ b/src/api/nodeRpcCall.js
@@ -3,13 +3,100 @@ import { USER_ROLES, userRolesHelper } from '../utils/userRolesHelper';
 import {handleMyDispatchErrors} from "../utils/therapist";
 import {newCompanyDataObject} from "../utils/defaultData";
 import * as centralizedBackend from './backend';
+import { BN_TWO } from '@polkadot/util';
+import { parseDollars, parseMerits } from '../utils/walletHelpers';
+import pako from 'pako';
+import { u8aToHex } from '@polkadot/util';
 
 const { ApiPromise, WsProvider } = require('@polkadot/api');
 
 const provider = new WsProvider(process.env.REACT_APP_NODE_ADDRESS);
 let __apiCache = null;
 const getApi = async () => {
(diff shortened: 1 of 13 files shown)


src/api/nodeRpcCall.js
```diff
@@ -913,52 +1012,60 @@ const getOfficialUserRegistryEntries = async (walletAddress) => {
...
-                pushObject['isEncrypted'] = dynamicFieldDataArray[keyIsEncrypted]
-                crossReferencedFieldDataArray.push(JSON.parse(JSON.stringify(pushObject)))
-              }
+    let companyData;
+    try {
+      const compressed = companyRegistryEntity.unwrap().data;
+      companyData = api.createType('CompanyData', pako.inflate(compressed)).toJSON();
+    } catch(e) {
+      console.error("Invalid company data", e); // FIXME actually display this to user somehow?
+      return;
```
MichalLustykNeti, line 1021: Shouldn't we return null?

src/api/nodeRpcCall.js
```diff
@@ -1088,20 +1168,7 @@ const getLandNFTMetadataJson = async (collection_id, nft_id) => {
...
-      },
-      "LandMetadata": {
-        "demarcation": "BoundedVec<Coords, u32>",
-        "type": "Text",
-        "status": "Text",
-      }
-    }
-  });
+  const api = await getApi();
   /*let metadata = {
```
MichalLustykNeti, line 1172: i believe we should clean up this

src/api/nodeRpcCall.js
kacperzuk-neti: Not sure it matters.

1. It should never happen
2. Both `undefined` (which I think will result in infinite loader) and `null` (which I think will result in "No registered companies") are wrong. I kinda thing `null` is worse, as it just lies to user, while `undefined` at least will clearly show as something that's broken. We just need a proper error handling for this.

I've created BLOCKCHAIN-254 ticket to implement the error handling. I don't want to modify this file, as there are other PRs that already rely on it, so I want to limit changes to critical only.

src/api/nodeRpcCall.js
kacperzuk-neti: Out of scope for this PR. And there are other PRs that depend on this one, so such change will cause more trouble than it's worth now.

src/components/Registries/RegistriesCompanies/CompanyCRUD/index.js
```diff
@@ -1,38 +1,50 @@
...
-    if(registryCRUDAction.action == 'edit'){
-      formObject = registryCRUDAction.dataObject
-      buttonMessage = 'Submit Company Edit'
-      companyId = registryCRUDAction.dataObject.id
-      callback = ((companyData) =>{})
+    if (registryCRUDAction.action === 'edit') {
+      formObject = registryCRUDAction.dataObject;
+      buttonMessage = 'Submit Company Edit';
+      companyId = registryCRUDAction.dataObject.id;
+      callback = ((_companyData) => {});
```
DorianSternVukotic, line 39: why the underscore ?

kacperzuk-neti: We're trying to make new/modified code have no errors when linted with eslint. One of the lints is that unused variables should be prefixed with `_`. `_companyData` here is unused.

kacperzuk-neti:
Approved on dev chat.

PR #170, opened 2023-11-21
BLOCKCHAIN-231 FE - chain explorer error handling is missing

MichalLustykNeti:


src/redux/reducers/bridge.js
```diff
@@ -144,8 +145,16 @@ const bridgeReducer = handleActions(
...
-      }
+        },
+      },
+    }),
+
+    [bridgeActions.getTransfersToEthereum.failure]: (state) => ({
+      ...state,
+      transfers: {
+        ...state.transfers,
+        toEthereumFailed: true,
```
kacperzuk-neti, line 156: We should reset it to `false` on `.call` - in case it starts working on next try.

kacperzuk-neti (changes requested):

kacperzuk-neti (approved):

PR #185, opened 2023-12-02
BLOCKCHAIN-294 FE Update staked amount after payour

MichalLustykNeti:


MichalLustykNeti:
> Title, branch, and code do not match.

My bad fall should be fixed now

DorianSternVukotic:
I dont really understand it now, I created the ticket when i payout rewards, then current stake didnt increase, i was assuming the solution would be to refresh balances after payout, but now its doing math with unstaking? How does this work ?

MichalLustykNeti:
> I dont really understand it now, I created the ticket when i payout rewards, then current stake didnt increase, i was assuming the solution would be to refresh balances after payout, but now its doing math with unstaking? How does this work ?

So unstake do not unstake immediately wait until some era to do so. We ALREADY reloaded the wallet balance after the scheduled withdrawal but it stays the same because no withdrawal was yet done it is only a plan to do so. That's why I added a math to it to show users that balance (that user may use/unstake) even after unstake schedule changes but on the blockchain, it is not until a specific era.

DorianSternVukotic:
> > I dont really understand it now, I created the ticket when i payout rewards, then current stake didnt increase, i was assuming the solution would be to refresh balances after payout, but now its doing math with unstaking? How does this work ?
> 
> So unstake do not unstake immediately wait until some era to do so. We ALREADY reloaded the wallet balance after the scheduled withdrawal but it stays the same because no withdrawal was yet done it is only a plan to do so. That's why I added a math to it to show users that balance (that user may use/unstake) even after unstake schedule changes but on the blockchain, it is not until a specific era.

"Payout" Refers specifically to payout of rewards, which is instant either into liquid or staked LLD pool. It is okey for unstaking not to be instant. Was this a misunderstanding of the task or just wrong title?

PR #191, opened 2023-12-07
Mobile responsiveness fixes

DorianSternVukotic:


MichalLustykNeti (changes requested):
Is it possible to use one of useMediaQuery or css @media? i believe when we will 4do it in one way it will be easier to maintain. From my perspective useMediaQuery is much better then css media.

DorianSternVukotic:
> Is it possible to use one of useMediaQuery or css @media? i believe when we will 4do it in one way it will be easier to maintain. From my perspective useMediaQuery is much better then css media.

I believe best use of useMediaQuery is when we need to display a totally different component with different logic, and css @media is for smaller visual only changes. For example, for navbar, which is massively different, i useMediaQuery and choose whether to use desktop or mobile component, but most other changes are css @media only

PR #192, opened 2023-12-07
BLOCKCHAIN-239 FE - add validation to all forms that are missing it

MichalLustykNeti:


src/components/Modals/SendLLDModal.js
```diff
@@ -4,19 +4,29 @@ import PropTypes from 'prop-types';
...
+import { walletSelectors } from '../../redux/selectors';
 
 function SendLLDModal({ closeModal }) {
   const dispatch = useDispatch();
+  const balances = useSelector(walletSelectors.selectorBalances);
+  const senderWallet = useSelector(walletSelectors.selectorWalletAddress);
+  const maxUnbond = BN.max(
+    BN_ZERO,
+    (new BN(balances?.liquidAmount?.amount ?? 0))
+      .sub(parseDollars('10')), // leave at least 10 liquid LLD...
```
DorianSternVukotic, line 27: Good catch that minimum should remain, however it shouldnt be 10 and the user should know that he cant send all. I will fix it though

src/components/Modals/SendLLDModal.js
```diff
@@ -46,18 +66,25 @@ function SendLLDModal({ closeModal }) {
         name="recipient"
         placeholder="Send to address"
         required
-        validate={(v) => isValidSubstrateAddress(v) || 'Invalid Address'}
+        validate={((v) => {
+          if (!isValidSubstrateAddress(v)) return 'Invalid Address';
+          if (v.toLowerCase() === senderWallet.toLowerCase()) return 'Invalid Address';
```
DorianSternVukotic, line 71: Its ok to send to yourself

PR #193, opened 2023-12-08
BLOCKCHAIN-297 - Discussion links per proposal hash + links for all referenda + multiple links per proposal

kacperzuk-neti:


src/api/nodeRpcCall.js
```diff
@@ -102,22 +102,20 @@ const getApi = async () => {
...
-    const centralizedData = allCentralizedData.find((cItem) => (
-      parseInt(cItem.chainIndex) == parseInt(item.index)
-    ))
-
     const proposalHash = isReferendum ? item.imageHash : (
       item.boundedCall?.Lookup?.hash_ ?? 
       item.boundedCall?.Legacy?.hash_
     )
+
+    const centralizedDatas = allCentralizedData.filter((cItem) => ( cItem.hash == proposalHash))
```
MichalLustykNeti, line 110: can we use '===' instead of '=='?

src/api/backend.js
```diff
@@ -57,10 +57,10 @@ export const getReferenda = async () => {
 };
 
 export const addReferendum = async ({
-  link, chainIndex, name, description, hash, additionalMetadata, proposerAddress,
+  link, name, description, hash, additionalMetadata, proposerAddress,
 }) => {
   const api = getApi();
-  return await api.post('/referenda', {
-    link, chainIndex, name, description, hash, additionalMetadata, proposerAddress,
+  return api.post('/referenda', {
```
MichalLustykNeti, line 63: can we add await here?

kacperzuk-neti: Why?

src/api/nodeRpcCall.js
kacperzuk-neti: Done.

PR #194, opened 2023-12-11
BLOCKCHAIN-289 FE Signup page cleanup

MichalLustykNeti:
image

### DIFF
--- a/src/components/AuthComponents/SignIn/index.js
+++ b/src/components/AuthComponents/SignIn/index.js
@@ -2,7 +2,7 @@
 import React, { useEffect } from 'react';
 import { useForm } from 'react-hook-form';
 import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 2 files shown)


kacperzuk-neti:
TBH the address select is currently meaningless as well. FE only cares about SSO and the address stored in centralized DB. I think it's safe to remove it as well.

@DorianSternVukotic thoughts?

DorianSternVukotic:
> TBH the address select is currently meaningless as well. FE only cares about SSO and the address stored in centralized DB. I think it's safe to remove it as well.
> 
> @DorianSternVukotic thoughts?

True, users will complain that it does nothing, BUT it being there is super useful for debugging... I have an idea, lets just give a list of available addresses will implement it

DorianSternVukotic:
@MichalLustykNeti dont put yarn lock in commits

PR #199, opened 2023-12-15
Blockchain 295 fe proposing referendums should mention it costs 100 lld

SzymonZiebaNeti:


src/components/Modals/AgreeDisagreeModal.js
```diff
@@ -0,0 +1,47 @@
...
+  onAgree,
+  onDisagree,
+  agreeButtonType,
+  children,
+  style,
+}) {
+  return (
+    <div className={classNames(styles.agreeDisagree, style)}>
+      <h3>{text}</h3>
+      {children || <span>This opperation cost 100 LLD.</span>}
```
DorianSternVukotic, line 18: costs instead  of cost

kacperzuk-neti, line 18:
```suggestion
      {children || <span>This operation costs 100 LLD.</span>}
```

2 typos

src/components/Modals/AgreeDisagreeModal.js
```diff
@@ -0,0 +1,47 @@
...
+      <div className={styles.buttons}>
+        <Button type="button" onClick={onDisagree} medium>
+          No
+        </Button>
+        <Button type={agreeButtonType} onClick={onAgree} primary medium>
+          Yes
```
kacperzuk-neti, line 24: I think we use Cancel and Submit instead of No and Yes in other places

src/components/Modals/CitizenRepealLegislationModal.js
```diff
@@ -32,94 +37,101 @@ function CitizenRepealLegislationModal({
...
 
+  const { dialogStep, handleClick } = useAgreeDisagreeModal(isValid, trigger);
+
   return (
     <form
       className={styles.getCitizenshipModal}
       onSubmit={handleSubmit(onSubmitRepeal)}
     >
-      <div className={styles.h3}>Propose referendum for legislation repeal</div>
+      {dialogStep === 'form' ? (
```
kacperzuk-neti, line 60: I dodnt test, but wont this cause whole form im the back to disappear from behind the modal? If yes then I dont like it, it will cause jarring jumps.

PR #202, opened 2023-12-18
BLOCKCHAIN-298 FE referendum 'proposal by' field should show identity…

SzymonZiebaNeti:
… name of the proposee


src/hooks/useUsersIdentity.js
```diff
@@ -0,0 +1,27 @@
...
+  const userOrId = userName || (proposer && truncate(proposer, 13));
+
+  useEffect(() => {
+    if (proposer) {
+      dispatch(identityActions.getIdentities.call(proposer));
+    }
+
+    if (listOfProposers) {
+      // eslint-disable-next-line max-len
+      listOfProposers.map((adress) => dispatch(identityActions.getIdentities.call(adress)));
```
DorianSternVukotic, line 20: This makes many calls, it should be a one call, i think we usually use query.multi with all identity queries

DorianSternVukotic (changes requested):

src/components/Voting/Referendum/Items/ProposalItem/index.js
```diff
@@ -1,6 +1,7 @@
 import React, { useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch, useSelector } from 'react-redux';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
kacperzuk-neti, line 4: Cant we remove the import?

src/components/Voting/Referendum/Items/ProposalItem/index.js
```diff
@@ -100,7 +104,7 @@ function ProposalItem({
             <div>
               <div className={styles.metaTextInfo}>
                 Proposed by:
-                <b>{ proposer && truncate(proposer, 13) }</b>
+                <b>{ findNameOrId(proposer, usersList) }</b>
```
kacperzuk-neti, line 107: How about instead of searching through the list for every entry, we preprocess it into a map and do direct accesses?

kacperzuk-neti (changes requested):

src/components/Voting/Referendum/Items/ProposalItem/index.js
SzymonZiebaNeti: Probably not, in next lines it's used

`  } else {
    // this sux but we have no other way until we refactor it to NOT use toJSON/toHuman
    hash = blake2AsHex(hexToU8a(boundedCall.Inline));
  }`

kacperzuk-neti: Then why `eslint-disable-next-line` is needed?

kacperzuk-neti: Ahh, I see. Cause we don't have it in package.json, but use it directly. Can we add it to package.json instead of sprinkling eslint disables?

SzymonZiebaNeti: yea, i will ad it

src/components/Voting/Referendum/Items/ReferendumItem/index.js
```diff
@@ -200,15 +201,19 @@ function ReferendumItem({
...
                       <li key={centralizedData.id}>
-                        <a href={sanitizeUrl} target="_blank" rel="noreferrer">{centralizedData.name}</a>
+                        <a href={sanitizeUrl} target="_blank" rel="noreferrer">
+                          {centralizedData.name}
+                        </a>
                         {' - '}
                         {centralizedData.description}
                         {' '}
-                        (Discussion added by
+                        (Discussion addede by
```
DorianSternVukotic, line 214: added

src/hooks/usersIdentity/useProposersList.js
```diff
@@ -0,0 +1,17 @@
+import { useMemo } from 'react';
+
+const useProposerList = (crossReferencedList, isProposer = false) => {
```
DorianSternVukotic, line 3: Why do this as a hook ?

src/api/nodeRpcCall.js
```diff
@@ -183,7 +183,7 @@ const getLldBalances = async (addresses) => {
   }
 };
 
-const getAdditionalAssets = async (address) => {
+const getAdditionalAssets = async (address, isIndexNeed = false) => {
```
DorianSternVukotic, line 186: I dont think this is required ? Some merge issue maybe ?

SzymonZiebaNeti: isIndexNeed is not required that's why i put there default value, 

there is one usage in getAdditionalAssetsWorker where it's used

`function* getAdditionalAssetsWorker() {
  const walletAddress = yield select(blockchainSelectors.userWalletAddressSelector);
  const additionalAssets = yield call(getAdditionalAssets, walletAddress);
  yield put(walletActions.getAdditionalAssets.success(additionalAssets));
}`

I know i can leave it undefined, but i want to keep type, because i worked with typescript and i think, that this is a good practice.

DorianSternVukotic: Okay but why is this part of "FE referendum 'proposal by' field should show identity" ?
I m not against merging small improvements like this, but you need to make it clear if its an unrelated small improvement or somehow part of the ticket

SzymonZiebaNeti: Okay, you mean, is this function required here, not parameter. So this is on develop right now, this function is changed to other task, so merge issue.

PR #206, opened 2023-12-28
BLOCKCHAIN-324 FE Popup info about using brave browser

SzymonZiebaNeti:


DorianSternVukotic:
Actually, this branch will be useful, but it should be part of the user guiding process, its not the only case we need to handle. For example, we should check if user is on mobile but not in a dApp explorer browser. This can maybe be checked with just seeing if wallet extensions are injected but maybe not. In any case, this task is finished, but i m not gonna merge it to dev, i ll copy paste the relevant code to my branch that checks for these things.

PR #207, opened 2023-12-28
BLOCKCHAIN-310 FE - user-provided URLs are insecure

SzymonZiebaNeti:


src/components/Voting/Referendum/Items/ProposalItem/index.js
```diff
@@ -117,25 +119,27 @@ function ProposalItem({
...
-                    />
-                    )
-                  </li>
-                ))}
+                {centralizedDatas.map((centralizedData) => {
+                  const html = jsxss(`<a href="${centralizedData.link}}">${centralizedData.name}</a>`);
+                  return (
+                    <li key={centralizedData.id}>
+                      {/* eslint-disable-next-line react/no-danger */}
+                      <div dangerouslySetInnerHTML={{ __html: html }} />
```
kacperzuk-neti, line 127: This approach kinda increases the attack surface. With JSX we have guarantees that attacker won't escape tags/attributes and the only problem is with URLs containing JS code.
By using this jsxss on raw HTML + dangerouslySetInnerHTML, we ditch the safety we've got from JSX. And I dont trust random libs enough to be fine with that.
How about using something more targeted, like https://www.npmjs.com/package/@braintree/sanitize-url ? Or, IMO even better, just whitelisting that the URL must begin with http:// or https://?

kacperzuk-neti (changes requested):

src/components/Voting/Referendum/Items/ProposalItem/index.js
SzymonZiebaNeti: I think, we should use this library with sanitize-url.  For user experience is better, because of links, when you copy some link from browser, you don't have http or https. So we should validate when it's come from user and tell them what's the problem or validate it, when we show it to user. But if we want to validate it, what about the polkadot js apps, who will be able, to add new proposal from there?

kacperzuk-neti: We can't rely on client-side validation. You can add it to make experience better for user, but ultimately we 100% need sanitization on display. Then we're safe even if someone adds nasty URL bypassing url (for example through the polkadot.js apps).

src/utils/sanitizeUrlHelper.js
```diff
@@ -0,0 +1,9 @@
+import { sanitizeUrl } from '@braintree/sanitize-url';
+
+const sanitizeUrlHelper = (url) => {
+  const sanitizedUrl = sanitizeUrl(url);
+  const checkerStartHttp = ['http', 'https'].some((element) => sanitizedUrl.includes(element));
```
kacperzuk-neti, line 5:
```suggestion
  const checkerStartHttp = ['http://', 'https://'].some((element) => sanitizedUrl.startsWith(element));
```

Otherwise it would allow urls like `http.im.evil.com` and `www.im.evil/http`.

kacperzuk-neti (approved):

PR #208, opened 2023-12-28
BLOCKCHAIN-308 FE legislation explorer propose menu empty

SzymonZiebaNeti:


src/components/Legislation/LegislationView/index.js
```diff
@@ -138,22 +143,23 @@ function ActionButtons({
...
-            )}
-            {tier !== 'Constitution' && (
-              <ProposeRepealLegislationButton {...{ tier, id, section }} />
-            )}
-            { (!repealReferendum && !repealProposal)
+        <div className={styles.dropdown}>
+          {isRepealOption && (
+          <RepealLegislationButton {...{ tier, id, section }} />
+          )}
+          {tier !== 'Constitution' && (
```
DorianSternVukotic, line 150: Gotta check if citizens can propose to repeal international treaty. If not, there would be a case when nothing is shown... I ll ask Ptacnik

DorianSternVukotic: i can confirm that citizens cannot start referendum to change international treaties. Only congress can. So add this condition please. This means that for citizens that arent congressmen, they shouldnt be able to propose anything regarding international treaties.

DorianSternVukotic (changes requested):

DorianSternVukotic:
As a side note, NOT a priority, we have duplicate logic of what is allowed for who and why in FE and Substrate, and we are currently hardcoding the checks. This is fine, but we can spend 30 minutes(but not 5 days :P ) thinking about if a better way exists. Might also be a question for Tomasz @kacperzuk-neti

src/components/Legislation/LegislationView/index.js
kacperzuk-neti: @DorianSternVukotic can you triple-check this with Ptacnik? Because [Referendum Law](https://github.com/liberland/laws/blob/Laws-Current/Referendum%20Law/referendum-law.md) Art. 1 Section 1 point 2 says:

```
To that end, the Citizens may, using the Referendum, enact, repeal and amend the following instruments of legislation:
[...]
b. The instrument of internal ratification for the international treaties of the highest importance, as stipulated by Art. 79 Section 2 of the Constitution.
```

Though it's true that the Art 79 Section 2 of the Constitution only mentions congress.

If indeed only Congress can propose treaty-changing referenda, then we need to:
1. Change Referendum Law (unless it's OK and I just don't understand legislative English well enough)
2. Change substrate ( this line: https://github.com/liberland/liberland_substrate/blob/develop/bin/node/runtime/src/lib.rs#L1449C1-L1449C59 )
3. Change frontend

kacperzuk-neti:
> As a side note, NOT a priority, we have duplicate logic of what is allowed for who and why in FE and Substrate, and we are currently hardcoding the checks. This is fine, but we can spend 30 minutes(but not 5 days :P ) thinking about if a better way exists. Might also be a question for Tomasz @kacperzuk-neti

There's no good way to automate it without making blockchain code very non-substratey

src/components/Legislation/LegislationView/index.js
DorianSternVukotic: Conclusion, "Lets not waste energy on this"

PR #209, opened 2024-01-02
BLOCKCHAIN-321 FE refactor duplicate api constructions

SzymonZiebaNeti:


DorianSternVukotic (approved):

kacperzuk-neti:
I know this is already merged, but I don't like it. Instead of exposing raw `axios` to globally, we should move `api.get('/users/me', ...` from auth component to `src/api/backend.js`.

PR #219, opened 2024-01-17
Blockchain 328 fe non blocking injection of wallets

SzymonZiebaNeti:


DorianSternVukotic:
This should also consider using web3AccountsSubscribe instead of web3Accounts if it doesnt already

kacperzuk-neti:
> This should also consider using web3AccountsSubscribe instead of web3Accounts if it doesnt already

In my tests `web3AccountsSubscribe` indeed seems to support multiple extensions perfectly.

kacperzuk-neti:
@DorianSternVukotic we reworked this a bit

1. sso-related auth and fetching user info from backend is handled by verifySession stuff
2. wallets/extension fetching is handled by subscribeWalletsSaga
3. GuidedSetup handles logic on whether everything is in order or not
4. the only thing we store outside redux is sso access token in localStorage - we might want to change it to a cookie with expiration date?. Either way, it's not in sessionStorage anymore, so I believe it solves BLOCKCHAIN-329 as a bonus
5. the initGetDataFromNode was problematic - I've dropped userWalletAddress from sessionStorage and keep it in redux only instead, but it only gets there after verifySession.success - initGetDataFromNode was launched too early. I've removed it and instead fetch data on-demand where it's needed. - this solves BLOCKCHAIN-341 as a bonus
6. while removing initGetDataFromNode I've notices half of these actions aren't really used - I've removed them
7. I've had an initial version that did await web3Enable and then web3AccountsSubscribe. But it turns out, it still sucked! Every once in a while, web3Enable would only detect subwallet, and while web3AccountsSubscribe correctly detected any changes in subwallet accounts, it never picked up polkadot-js. So we're back to a loop... but now we do it with subscriptions & dynamically, so we can just patiently wait for it to show up without delaying app startup.

Summary: solves 3 tickets AND reduces code by ~30 lines.

src/components/GuidedSetup/NoConnectedWalletComponent/index.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import WalletListComponent from '../WalletListComponent';
+import { proposWalletListUserID } from '../propsTypes/propTypes';
```
DorianSternVukotic, line 3: typo

PR #227, opened 2024-01-23
BLOCKCHAIN-328 rewrite

kacperzuk-neti:
@SzymonZiebaNeti lets discuss this on call tomorrow

tl;dr:
1. sso-related auth and fetching user info from backend is handled by verifySession stuff
2. wallets/extension fetching is handled by subscribeWalletsSaga
3. GuidedSetup handles logic on whether everything is in order or not
4. the only thing we store outside redux is sso access token in localStorage - we might want to change it to a cookie with expiration date?
5. the `initGetDataFromNode` was problematic - I've dropped userWalletAddress from sessionStorage and keep it in redux only instead, but it only gets there after `verifySession.success` - `initGetDataFromNode` was launched too early. I've removed it and instead fetch data on-demand where it's needed.
6. while removing `initGetDataFromNode` I've notices half of these actions aren't really used - I've removed them
7. I've had an initial version that did `await web3Enable` and then `web3AccountsSubscribe`. But it turns out, it still sucked! Every once in a while, `web3Enable` would only detect subwallet, and while `web3AccountsSubscribe` correctly detected any changes in subwallet accounts, it never picked up polkadot-js. So we're back to a loop... but now we do it with subscriptions & dynamically, so we can just patiently wait for it to show up without delaying app startup.

Draft as I still need to review my own diff first, but it's already late today :/


src/components/GuidedSetup/index.js
```diff
@@ -1,47 +1,100 @@
...
+
+  if (extensions.length === 0) {
+    return (
+      <GuidedSetupWrapper>
+        <NoWalletsDetectedInBrowser />
+      </GuidedSetupWrapper>
+    );
+  }
+
+  if (wallets.length === 0) {
```
kacperzuk-neti, line 73: Opps, I think I misunderstood this one. This should be checking if `userWalletAddress` is empty.

SzymonZiebaNeti: This should be with extensions check, if there is no extension, or any wallets, it's should show NoWalletDetectedInBrowser component.
But you are right, if userWalletAdress is empty it's should show NoConnectedWalletComponent.

src/components/Voting/CongressionalAssemble/index.js
```diff
@@ -1,17 +1,23 @@
...
 function CongressionalAssemble() {
+  const dispatch = useDispatch();
   const democracy = useSelector(democracySelectors.selectorDemocracyInfo);
   const [selectedCandidates, setSelectedCandidates] = useState([]);
   const [eligibleUnselectedCandidates, setEligibleUnselectedCandidates] = useState([]);
   const [didChangeSelectedCandidates, setDidChangeSelectedCandidates] = useState(false);
 
+  useEffect(() => {
+    dispatch(democracyActions.getDemocracy.call());
+  }, []);
```
SzymonZiebaNeti, line 19: We don't want 'dispatch' in dependencies?

kacperzuk-neti: We do want it in deps, I've forgot it - nice catch!
Shouldn't eslint complain about it...?

SzymonZiebaNeti: We need to add rules for this, in our eslint config we don't have it. I can do it, if we want this.

SzymonZiebaNeti (approved):

PR #228, opened 2024-01-24
Add eslint rules for react hooks

SzymonZiebaNeti:


package.json
```diff
@@ -25,6 +25,7 @@
     "crypto-browserify": "^3.12.0",
     "css-loader": "^6.7.1",
     "dotenv-webpack": "^7.1.0",
+    "eslint-plugin-react-hooks": "^4.6.0",
```
DorianSternVukotic, line 28: shouldnt this be a dev dependency ?

DorianSternVukotic (changes requested):

package.json
SzymonZiebaNeti: In my opinion we should have it, it's better to show warning all the time, when someone forgot to put for example, 
 something in useEffect dependencies, this will warn about it. When you are sure, you don't want some variables in dependencies, then we can disable eslint.
 This changes comes after talk with Kacper, when i did review i found out, that there can be missing dispatch in dependencies. So this will prevent, asking, if someone want to miss something in dependencies (if you need to put eslint disable you need to think about it), if not, eslint will talk about it in vsc.

PR #234, opened 2024-02-07
BLOCKCHAIN-359 FE allow swapping wallets while logged in

SzymonZiebaNeti:


DorianSternVukotic:
I like it!

Just one more feature request.

When using a different wallet address, add a button "Switch to registered wallet", that way its easier for the user to understand why there are no buttons anymore.

DorianSternVukotic:
And also, lets allow staking with multiple accounts

PR #246, opened 2024-02-21
Blockchain 379 fe company registry certificate

SzymonZiebaNeti:
Merge after

https://liberland.atlassian.net/jira/software/projects/BLOCKCHAIN/boards/2?selectedIssue=BLOCKCHAIN-380

https://github.com/liberland/liberland-blockchain-middleware/pull/1


src/api/middleware.js
```diff
@@ -7,5 +7,26 @@ const getMiddlewareApi = () => axios.create({
...
+export const generatePdf = async (usingWalletAddress, companyId, pathName, blockNumber) => {
+  const userToken = JSON.parse(localStorage.getItem('ROCP_token'));
+  const middlewareApi = getMiddlewareApi();
+  try {
+    const response = await middlewareApi.post('/v1/certificate', {
+      usingWalletAddress,
+      userToken,
+      companyId,
+      pathName,
+      blockNumber,
```
DorianSternVukotic, line 24: you dont need user token in fact middleware only uses

 	const { companyId, pathName, blockNumber } = req.body;

src/components/Registries/RegistriesCompanies/index.js
```diff
@@ -13,34 +14,54 @@ import styles from './styles.module.scss';
...
+
+    const href = URL.createObjectURL(blob);
+    const file = `${pathName}.pdf`;
+    const link = document.createElement('a');
+    link.href = href;
+    link.setAttribute('download', file);
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
+    URL.revokeObjectURL(href);
```
DorianSternVukotic, line 62: what is this and why is it like this

DorianSternVukotic (changes requested):

src/components/Registries/RegistriesCompanies/index.js
SzymonZiebaNeti: What's wrong with it?

get blob from middleware,
creates a unique URL for the Blob, 
create new element - anchor for the download link,
set href as our url blob,
add download attribute for browser, and set name of download file,
add to dom,
initiating the download process,
clear code

PR #249, opened 2024-02-25
BLOCKCHAIN-357 FE company browser

SzymonZiebaNeti:


src/api/nodeRpcCall.js
```diff
@@ -964,6 +964,34 @@ const getLegislation = async (tier) => {
   return legislationById;
 };
 
+const getOfficialRegistryEntries = async () => {
```
DorianSternVukotic, line 967: I think we should split this.

GetCompanyRegistryEntries should be most of the logic
GetOfficialCompanyRegistryEntries should call the GetCompanyRegistryEntries function and pass it a 0 - the index of official registry

PR #253, opened 2024-03-04
BLOCKCHAIN-403 Docs and functions Simple contracts

MichalLustykNeti:


src/api/nodeRpcCall.js
```diff
@@ -1924,6 +1924,63 @@ const fetchPendingIdentities = async () => {
...
+
+const getAllContracts = async () => {
+  const api = await getApi();
+
+  const [rawJudgesSignatures, rawPartiesSignatures] = await Promise.all([
+    api.query.contractsRegistry.judgesSignatures.entries(),
+    api.query.contractsRegistry.partiesSignatures.entries(),
+  ]);
+
+  const rawContracts = await api.query.contractsRegistry.contracts.entries();
```
kacperzuk-neti, line 1939: Use queryMulti instead: https://polkadot.js.org/docs/api/start/api.query.multi/#multi-queries-distinct-types
If queryMulti doesn't work with `.entries` (I'm not sure TBH), put `rawContracts` in the Promise.all as well.

src/api/nodeRpcCall.js
```diff
@@ -1924,6 +1924,63 @@ const fetchPendingIdentities = async () => {
   return processed.filter((entity) => entity.data.judgements.length === 0);
 };
 
+const getSignaturesForGivenId = async (allSignatures, id) => allSignatures.find(
```
kacperzuk-neti, line 1927: This shouldn't be `async`.

src/api/nodeRpcCall.js
```diff
@@ -1924,6 +1924,63 @@ const fetchPendingIdentities = async () => {
...
+const getAllContracts = async () => {
+  const api = await getApi();
+
+  const [rawJudgesSignatures, rawPartiesSignatures] = await Promise.all([
+    api.query.contractsRegistry.judgesSignatures.entries(),
+    api.query.contractsRegistry.partiesSignatures.entries(),
+  ]);
+
+  const rawContracts = await api.query.contractsRegistry.contracts.entries();
+  return Promise.all(rawContracts.map(async ([id, contract]) => {
```
kacperzuk-neti, line 1940: There's nothing async here except for `getSignaturesForGivenId` which shouldn't be async.. So no need for `async` nor Promise.all here.

src/api/nodeRpcCall.js
```diff
@@ -1924,6 +1924,63 @@ const fetchPendingIdentities = async () => {
...
+    const contractId = id.args[0];
+    const [judgesSignatures, partiesSignatures] = await Promise.all([
+      getSignaturesForGivenId(rawJudgesSignatures, contractId),
+      getSignaturesForGivenId(rawPartiesSignatures, contractId),
+    ]);
+
+    const shouldBeData = contract.unwrapOr(null);
```
kacperzuk-neti, line 1947: Call current contract `maybeContract` and then:
```suggestion
    const contract = maybeContract.unwrapOr(null);
```

src/api/nodeRpcCall.js
```diff
@@ -1924,6 +1924,63 @@ const fetchPendingIdentities = async () => {
...
+    return {
+      contractId,
+      data: shouldBeData ? shouldBeData.data : null,
```
kacperzuk-neti, line 1950: After doing the `shouldBeData` -> `contract` rename from previous comment you can use this instead of ternary operator:

```suggestion
      data: contract?.data,
```

And same for parties/creator/deposit.

src/api/nodeRpcCall.js
```diff
@@ -1924,6 +1924,63 @@ const fetchPendingIdentities = async () => {
...
+const removeContract = async (contractId, walletAddress) => {
+  const api = await getApi();
+  const extrinsic = api.tx.contractsRegistry.removeContract(contractId);
+  return submitExtrinsic(extrinsic, walletAddress, api);
+};
+
+const getAllJudges = async () => {
+  const api = await getApi();
+  const rawJudges = await api.query.contractsRegistry.judges.entries();
+  return rawJudges.map(([address, isJudge]) => isJudge.isTrue && address.args[0]).filter((address) => address);
```
kacperzuk-neti, line 1981: That's a weird order. Either swap filter and map:
```suggestion
  return rawJudges.filter(([, isJudge]) => isJudge.isTrue).map(([address, ]) => address.args[0]);
```
Or use reduce.

kacperzuk-neti (changes requested):

src/api/nodeRpcCall.js
MichalLustykNeti: queryMulti doesn't work with .entries

kacperzuk-neti (approved):

PR #254, opened 2024-03-06
BLOCKCHAIN-387 FE DEX 

MichalLustykNeti:


src/api/nodeRpcCall.js
```diff
@@ -1924,6 +1997,115 @@ const fetchPendingIdentities = async () => {
...
+  lpTokenBurn,
+  amount1MinReceive,
+  amount2MinReceive,
+  withdrawTo,
+  walletAddress,
+) => {
+  const api = await getApi();
+  const extrinsic = api.tx.assetConversion.removeLiquidity(
+    asset1,
+
```
kacperzuk-neti, line 2094:
```suggestion
```

kacperzuk-neti (changes requested):

kacperzuk-neti (approved):

MichalLustykNeti:
image
If pool exist but there are no liquidity in it

MichalLustykNeti:
for 
asset1=1000000000000
asset2=1000000000000
image
reserves shouldn't be zero

MichalLustykNeti:
if there are not enough funds in reserves we should let user know how many token max he may swap. For now i have to guess how many funds can i swap
image

src/components/Wallet/Exchange/ExchangeShowMore/index.js
```diff
@@ -0,0 +1,71 @@
...
+        <div>
+          {`In the pool
+              ${formatProperlyValue(asset1, reserved.asset1, asset1ToShow, asset1Decimals || 0)}  
+            / ${formatProperlyValue(asset2, reserved.asset2, asset2ToShow, asset2Decimals || 0)}` }
+        </div>
+        )}
+      </div>
+      <div className={styles.liquidity}>
+        {liquidity
+          ? `Your liquidity: ${((liquidity / lpTokensBalance) * 100)}% (${liquidity} Lp Tokens)`
```
kacperzuk-neti, line 38: % isn't really important. In a healthy system, all users will have some very small share of the LP.
Number of LP tokens similarly isn't really useful to a user.
What is useful is how many asset1 and asset2 "belongs" to the user.

See https://support.uniswap.org/hc/article_attachments/21158962149645 for an example of how it looks in uniswap v2, on which our dex is loosely based.

src/api/nodeRpcCall.js
```diff
@@ -1925,6 +2029,191 @@ const fetchPendingIdentities = async () => {
...
+const getDexPools = async (walletAddress) => {
+  try {
+    const api = await getApi();
+    const pools = await api.query.assetConversion.pools.entries();
+    const poolsData = await Promise.all(pools.map(async ([poolKey, maybePoolData]) => {
+      const [asset1, asset2] = poolKey.args[0];
+      const { lpToken } = maybePoolData.unwrapOrDefault();
+      const asset1checkIsNative = asset1.value.toString();
+      const asset2checkIsNative = asset2.value.toString();
+      const asset2Tronsform = asset2checkIsNative || asset2.toString();
```
kacperzuk-neti, line 2095:
```suggestion
      const asset2Transform = asset2checkIsNative || asset2.toString();
```
This typo shows up in other places too, please fix all.

src/components/Wallet/Exchange/ExchangeShowMore/index.js
kacperzuk-neti, line 38: I have trouble tracing what `lpTokensBalance` is. However, we should divide by any balance here - we should divide by total supply of these tokens.

kacperzuk-neti (changes requested):

DorianSternVukotic:
Conflicts

src/redux/reducers/dex.js
```diff
@@ -0,0 +1,59 @@
...
+    }),
+
+    [dexActions.getPools.success]: (state, action) => ({
+      ...state,
+      pools: action.payload,
+    }),
+
+    [dexActions.getDexReserves.success]: (state, action) => ({
+      ...state,
+      reserves: { [action.payload.asset1Number + action.payload.asset2Number]: action.payload, ...state.reserves },
```
kacperzuk-neti, line 52: Won't this result in conflicts? What if we have pool for asset numbers `1` and `4` and a pool for assets `2` and `3`? They both will sum to key `5`.

src/utils/dexFormater.js
```diff
@@ -0,0 +1,282 @@
...
+  const amount2Min = calculateAmountMin(
+    amount2,
+    minAmountPercent,
+  );
+  return {
+    amount1, amount2, amount1Min, amount2Min,
+  };
+};
+
+export const converTransferData = async (
```
kacperzuk-neti, line 143:
```suggestion
export const convertTransferData = async (
```

kacperzuk-neti (changes requested):

src/redux/reducers/dex.js
SzymonZiebaNeti: asset1Number should be string, I want to make strings 14 and 23, but i will check if there is always string

kacperzuk-neti: That doesnt change it. If asset1 is `1` and asset2 is `21` it will conflict with assets `12` and `1`

src/utils/dexFormater.js
```diff
@@ -0,0 +1,282 @@
...
+  const stringValue = mod.toString();
+
+  const integerPart = stringValue.slice(0, -1e18.toString().length) || '0';
+  const decimalPart = stringValue.slice(-1e18.toString().length);
+
+  const dividedValueString = `${integerPart}.${decimalPart}`;
+  const substract = new BN(Number(div) + Number(dividedValueString));
+  const result = subResult
+    ? amount.sub(substract.isZero() ? new BN(2) : substract)
+    : amount.add(substract.isZero() ? new BN(2) : substract);
```
kacperzuk-neti, line 100: What this code does?

src/utils/dexFormater.js
```diff
@@ -0,0 +1,282 @@
...
+  return result;
+};
+
+export const calculateAmountDesiredFormatted = (desiredAmount, decimals) => {
+  const valueString = desiredAmount.toString();
+  const decimalsFromValue = valueString.split('.')[1]?.length || '0';
+  return new BN(valueString.replace('.', ''))
+    .mul(decimals)
+    .div((new BN(10)
+      .pow(new BN(decimalsFromValue))));
```
kacperzuk-neti, line 110: Don't reinvent parsing balances, use `parseAssets` from walletHelpers.

kacperzuk-neti (changes requested):
Parsing/formatting balances (decimal handling) logic needs to be refactored.
1. Most of it should be done using `parseAssets` and `formatAssets` from `walletHelpers`
2. For others there's still way to many places that do random `replace('.', '')`, `split('.')` or `pow(decimals)`. Rethink this, find common abstractions and refactor. Right now I guarantee that there's a dozen different subtle bugs in there.

src/components/Modals/TradeTokens.js
```diff
@@ -0,0 +1,269 @@
...
+  }, [asset1, asset2, reserves]);
+
+  const {
+    handleSubmit,
+    register,
+    formState: { errors, isValid },
+  } = useForm({
+    mode: 'onChange',
+    defaultValues: {
+      cos: null,
```
kacperzuk-neti, line 67: Some old placeholder?

src/components/Modals/TradeTokens.js
```diff
@@ -0,0 +1,269 @@
...
+      );
+      const swapData = {
+        path: { asset1, asset2 },
+        amount,
+        amountMin,
+        sendTo: walletAddress,
+      };
+      dispatch(
+        isBuy
+          ? dexActions.swapTokensForExactTokens.call(swapData) : dexActions.swapExactTokensForTokens.call(swapData),
```
kacperzuk-neti, line 93: I think there might be some misunderstanding here.
Both this calls do the same kind of trade - sell asset1 in exchange for asset2. It's just that with `swapTokensForExactTokens` is then you're saying "I sell asset1, I get 100 asset2" and `swapExactTokensForTokens` is "I sell 100 asset1, I get asset2". To actually sell asset2 and buy asset1 you need to swap them in the `path`.

Right now UI shows only "Buy LLD for Asset" and "Sell LLD For Asset". This isn't enough. In general I think that using custom buttons for this plainly sux UX-wise and even if us developers get confused our users have no chance and we should just copy Uniswap, buuut if we want to keep buttons, then we need 4 of them (there's a Polish CEX (zonda) that kinda does it like that, though with tabs instead of buttons):
* `Buy LLD for Asset` - swapTokensForExactTokens, where Asset is asset1 and LLD is asset2
* `Buy Asset for LLD` - swapTokensForExactTokens, where LLD is asset1 and Asset is asset2
* `Sell LLD for Asset` - swapExactTokensForTokens, where LLD is asset1 and Asset is asset2
* `Sell Asset for LLD` - swapExactTokensForTokens, where Asset is asset1 and LLD is asset2

In other words, `asset1` is always what user will lose, `asset2` is what user will gain.  Whether we use swapTokensForExactTokens or swapExactTokensForTokens depends on which amount user wants to specify, but doesn't affect the direction of the trade.

@DorianSternVukotic this is why you can't fix the ratio.

kacperzuk-neti (changes requested):

src/components/Modals/TradeTokens.js
kacperzuk: After signal chat, we keep 2 buttons "Buy LLD for Asset" and "Sell LLD for Asset" and they decide the direction of trade (which asset is asset1 and which is asset2). For the forms, we do two inputs: one for how much we pay (asset1) and one for how much we receive (asset2). Then, depending on which input the user puts amount in, we decide on which side is exact (pay or receive). So:
1. `Buy LLD for Asset` - Asset is asset1, LLD is asset2
2. `Sell LLD for Asset` - LLD is asset1, Asset is asset2
3. If user puts amount in the first input (the "I lose/pay with" input), we do `swapExactTokensForTokens`
4. If user puts amount in the second input (the "I buy/receive"), we do `swapTokensForExactTokens`

src/components/Modals/AddLiquidityModal.js
```diff
@@ -87,55 +78,41 @@ function AddLiquidityModal({
...
+    const sanitizedValue = sanitizeValue(value.toString());
+    const assetFormat = isAsset1 ? asset2 : asset1;
+    const amount = parseProperlyValue(
+      isAsset1 ? asset1 : asset2,
+      sanitizedValue,
+      isAsset1 ? assetData1?.decimals : assetData2?.decimals,
+    );
+    const tradeData = await (isAsset1
+      ? getSwapPriceExactTokensForTokens
+      : getSwapPriceTokensForExactTokens)(enum1, enum2, amount, false);
```
kacperzuk-neti, line 107:
```suggestion
    const getSwapPrice = isAsset1 ? getSwapPriceExactTokensForTokens : getSwapPriceTokensForExactTokens;
    const tradeData = await getSwapPrice(enum1, enum2, amount, false);
```
Will be much more readable.

src/utils/dexFormater.js
```diff
@@ -55,21 +60,33 @@ export const makeAssetToShow = (asset, symbol) => {
...
   if (asset === 'Native') {
-    formattedValue = formatDollars(value);
+    formattedValue = formatDollars(value, true);
   } else if (asset === '1') {
-    formattedValue = formatMerits(value);
+    formattedValue = formatMerits(value, true);
   } else {
-    formattedValue = formatter(value, decimals);
+    formattedValue = formatAssets(value, decimals || 0, true);
   }
```
kacperzuk-neti: If we have decimals, then we can just use `formatAssets` for everything.

src/utils/dexFormater.js
```diff
@@ -55,21 +60,33 @@ export const makeAssetToShow = (asset, symbol) => {
   return assetToShow;
 };
 
-export const formatProperlyValue = (asset, value, symbol, decimals) => {
+export const formatProperlyValue = (asset, value, decimals, symbol) => {
```
kacperzuk-neti, line 63: Terrible name. Maybe `formatAssetsWithSymbol` and move to `walletHelpers`? Or even better. Change the formatAssets in walletHelpers to optionally accept symbol and add it if provided.

src/utils/dexFormater.js
```diff
@@ -55,21 +60,33 @@ export const makeAssetToShow = (asset, symbol) => {
...
 
+export const parseProperlyValue = (asset, value, decimals) => {
+  let parsedValue;
+  if (asset === 'Native') {
+    parsedValue = parseDollars(value);
+  } else if (asset === '1') {
+    parsedValue = parseMerits(value);
+  } else {
+    parsedValue = parseAssets(value, decimals);
+  }
```
kacperzuk-neti, line 85: If we have decimals, then we can just call `parseAssets(value, decimals)` even for LLD and LLM, which makes `parseProperlyValue` totally redundant.

src/utils/dexFormater.js
```diff
@@ -85,47 +102,30 @@ const calculateAmountMin = (
...
-  const decimalsFromValue = valueString.split('.')[1]?.length || '0';
-  return new BN(valueString.replace('.', ''))
-    .mul(decimals)
-    .div((new BN(10)
-      .pow(new BN(decimalsFromValue))));
+  const minAmountValue = (Number(minAmountPercent || defaultMinPercent) / 100).toString();
+  const lengthDecimalPercent = minAmountValue.split('.')[1].length;
+  const minAmountPercentBN = parseAssets(minAmountValue, lengthDecimalPercent);
+  const decimalsBN = new BN(10).pow(new BN(lengthDecimalPercent));
+  const percentBN = amount.mul(minAmountPercentBN).divn(decimalsBN);
```
kacperzuk-neti, line 109: Way too overengineered. Just assume a big enough denominator instead of trying to guess it based on string formatting of the fraction. So something like this:
```
const denominator = 10000;
const slippagePercentBN = new BN(minAmountPercent * denominator / 100);
const maxSlippage = amount.mul(slippagePercentBN).div(denominator);
```

src/utils/dexFormater.js
```diff
@@ -85,47 +102,30 @@ const calculateAmountMin = (
...
+
+  if (subResult) {
+    const sub = amount.sub(percentBN.isZero() ? new BN(3) : percentBN);
+    return sub.lte(new BN(0)) ? 1 : sub;
+  }
+  return amount.add(percentBN.isZero() ? new BN(3) : percentBN);
```
kacperzuk-neti, line 115: I assume that the `add` case is actually for calculating `amountMax`, not `amountMin`. This doesn't belong in a single function. Call this function (that calculates the slippage) `calculateSlippage` and do a separate `calculateAmountMin` that subs it and `calculateAmountMax` that adds it.

Also it's not a formatter in any way, it's business logic, so doesn't belong in `dexFormater.js`

src/utils/dexFormater.js
```diff
@@ -1,7 +1,12 @@
 import { formatBalance, BN } from '@polkadot/util';
```
kacperzuk-neti, line 1: Typo in filename - should be `dexFormatter.js`

src/components/Modals/AddLiquidityModal.js
kacperzuk-neti: The same pattern is repeated in multiple places btw.

kacperzuk-neti (changes requested):

src/api/nodeRpcCall.js
```diff
@@ -5,9 +5,8 @@ import { USER_ROLES, userRolesHelper } from '../utils/userRolesHelper';
 import { handleMyDispatchErrors } from '../utils/therapist';
 import { blockchainDataToFormObject } from '../utils/registryFormBuilder';
 import * as centralizedBackend from './backend';
-import {
-  convertAssetData,
-} from '../utils/dexFormater';
+// eslint-disable-next-line import/no-cycle
```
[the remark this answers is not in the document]
SzymonZiebaNeti: It's because in dexFormatter we have import from nodeRpcCall, and in nodeRpcCall i have import from dexFormatter? should i just move this function convertAssetData to nodeRpcCall?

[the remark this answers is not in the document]
kacperzuk-neti: In perfect world I'd keep `nodeRpcCall` as a very thin wrapper over actual APIs and not put business logic in here at all. `convertAssetData` is something that fits more in sagas.

But I don't care much, there's a lot of business logic in here either way. I just missed the import of `nodeRpcCall` in `dexFormatter` and was genuinely confused as to why we need to silence eslint here. If it's a real cycle and it causes no issues I'm fine leaving it as-is.

kacperzuk-neti (changes requested):

kacperzuk-neti (approved):

kacperzuk-neti (changes requested):

src/components/Modals/AddLiquidityModal.js
```diff
@@ -0,0 +1,243 @@
...
+    const formatedValue = formatAssets(tradeData, decimalsOut);
+    const sanitizedValue = sanitizeValue(formatedValue.toString());
+    setValue('amount1Desired', isAsset1 ? value : sanitizedValue);
+    setValue('amount2Desired', isAsset1 ? sanitizedValue : value);
+    await trigger();
+  };
+
+  useEffect(() => {
+    dispatch(walletActions.getAssetsBalance.call([asset1, asset2]));
+  }, [dispatch, asset1, asset2]);
```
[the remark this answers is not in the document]
kacperzuk-neti: Not needed, this is modal.

src/components/Modals/TradeTokens.js
```diff
@@ -0,0 +1,354 @@
...
+    return true;
+  };
+
+  useEffect(() => {
+    dispatch(dexActions.getDexReserves.call({ asset1, asset2 }));
+  }, [dispatch, asset1, asset2]);
+
+  useEffect(() => {
+    dispatch(walletActions.getAssetsBalance.call([asset1, asset2]));
+  }, [dispatch, asset1, asset2]);
```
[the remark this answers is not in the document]
kacperzuk-neti: Not needed, this is modal.

src/components/Modals/RemoveLiquidity/index.js
```diff
@@ -46,14 +47,14 @@ function RemoveLiquidityModal({
...
+    const tokensToBurn = new BN(lpTokensBalance).mul(new BN(numberValue)).div(BN_HUNDRED);
     setTokensToBurnState(tokensToBurn);
-    const fee = 1; // 0.1%
+    const fee = new BN(withdrawlFee);
     const calculatedAmount1 = calculatePooled(tokensToBurn, liquidity, reserved.asset1);
     const calculatedAmount2 = calculatePooled(tokensToBurn, liquidity, reserved.asset2);
-    const asset1Data = calculatedAmount1.mul(new BN(BN_THOUSAND - fee)).div(new BN(BN_THOUSAND));
-    const asset2Data = calculatedAmount2.mul(new BN(BN_THOUSAND - fee)).div(new BN(BN_THOUSAND));
+    const asset1Data = !fee.isZero() ? calculatedAmount1.sub(calculatedAmount1.div(fee)) : calculatedAmount1;
+    const asset2Data = !fee.isZero() ? calculatedAmount2.sub(calculatedAmount2.div(fee)) : calculatedAmount2;
```
kacperzuk-neti, line 57: Why `amount - amount/fee`? Shouldn't it be `amount * (1000-fee)/1000` or `amount - amount * fee/1000`?

src/redux/actions/dex.js
```diff
@@ -45,4 +46,9 @@ export const {
     SUCCESS: undefined,
     FAILURE: undefined,
   },
+  GET_WITHDRAWL_FEE: {
```
kacperzuk-neti, line 49:
```suggestion
  GET_WITHDRAWAL_FEE: {
```
Typo.

kacperzuk-neti (changes requested):

src/components/Modals/RemoveLiquidity/index.js
```diff
@@ -0,0 +1,196 @@
...
+    setTokensToBurnState(tokensToBurn);
+    const fee = new BN(withdrawalFee);
+    const calculatedAmount1 = calculatePooled(tokensToBurn, liquidity, reserved.asset1);
+    const calculatedAmount2 = calculatePooled(tokensToBurn, liquidity, reserved.asset2);
+    const asset1Data = !fee.isZero()
+      ? calculatedAmount1.sub(calculatedAmount1.mul(fee).div(BN_MILLION))
+      : calculatedAmount1;
+    const asset2Data = !fee.isZero()
+      ? calculatedAmount2.sub(calculatedAmount2.mul(fee).div(BN_MILLION))
+      : calculatedAmount2;
```
kacperzuk-neti, line 61:
```suggestion
    const asset1Data = calculatedAmount1.sub(calculatedAmount1.mul(fee).div(BN_MILLION));
    const asset2Data = calculatedAmount2.sub(calculatedAmount2.mul(fee).div(BN_MILLION));
```
Multiplication will handle `fee = 0` just fine

kacperzuk-neti (approved):

DorianSternVukotic (changes requested):

PR #257, opened 2024-03-08
add automated ICM tokens reading and writing to identity office

DorianSternVukotic:
To add ICM reading and writing of merits and dollars, we need admin 2fa login with different sso client and context.

This is surprisingly not straightforward as the library does not support this. The solution Varun and me came up with is to use another login link which will set the authconfig which will be kept in localstorage so it persists after sso redirects.

The rest is logic to read ICM merits and dollars and write the corresponding, but negative transaction once an admin onboards an user using the offices identityoffice interface

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -15,6 +15,18 @@ const etherConfig = {
     [process.env.REACT_APP_ETHER_CHAIN_ID]: process.env.REACT_APP_ETHER_PROVIDER,
   },
 };
+const adminAuthConfig = {
+  clientId: `${process.env.REACT_APP_SSO_API_ADMIN_CLIENT_ID}`,
+  authorizationEndpoint: `${process.env.REACT_APP_SSO_API}/oauth/authorize`,
+  tokenEndpoint: `${process.env.REACT_APP_SSO_API}/oauth/token`,
+  redirectUri: `${process.env.REACT_APP_SSO_API_ADMIN_LINK}`,
+  scope: 'others:read_write',
+  postLogin: () => {
+    store.dispatch(authActions.verifySession.call());
+    store.dispatch(onBoardingActions.getEligibleForComplimentaryLld.call());
+  },
+  decodeToken: false,
+};
 
 const authConfig = {
   clientId: `${process.env.REACT_APP_SSO_API_CLIENT_ID}`,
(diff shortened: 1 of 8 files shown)


src/index.js
```diff
@@ -15,6 +15,18 @@ const etherConfig = {
     [process.env.REACT_APP_ETHER_CHAIN_ID]: process.env.REACT_APP_ETHER_PROVIDER,
   },
 };
+const adminAuthConfig = {
+  clientId: `${process.env.REACT_APP_SSO_API_ADMIN_CLIENT_ID}`,
+  authorizationEndpoint: `${process.env.REACT_APP_SSO_API}/oauth/authorize`,
+  tokenEndpoint: `${process.env.REACT_APP_SSO_API}/oauth/token`,
+  redirectUri: `${process.env.REACT_APP_FRONTEND_REDIRECT}/admin=true`,
```
MichalLustykNeti, line 22: This url was created in the same way in few files we should avoid that an use some function that would create it instead. Thanks to that in future if we will have any update to "creation" of this url we simple change it in one file instead of searching it in whole project. (I believe the same approach would be nice do be done for the rest of does links as well)

src/index.js
```diff
@@ -28,10 +40,20 @@ const authConfig = {
...
 };
 
+const queryString = window.location.search;
+const urlParams = new URLSearchParams(queryString);
+const isAdminLogin = urlParams.get('admin')
+if(isAdminLogin === 'true') {localStorage.setItem('isAdminLogin', 'true')}
+
+const useAuthConfig = localStorage.getItem('isAdminLogin') === 'true' ? adminAuthConfig : authConfig
+
+console.log('useAuthConfig')
```
MichalLustykNeti, line 50: leftover

src/index.js
```diff
@@ -28,10 +40,20 @@ const authConfig = {
...
+console.log(useAuthConfig)
```
MichalLustykNeti, line 51: leftover

MichalLustykNeti (changes requested):

MichalLustykNeti:
Also is eslint happy?

DorianSternVukotic:
> Also is eslint happy?

No but personally i dont care for eslint

src/index.js
DorianSternVukotic: I agree with the admin link, this is a special case that needs to be handled, for the other ones, its ok as is, because this is the only place we should ever have these links in.

PR #262, opened 2024-03-18
BLOCKCHAIN-413 FE unstaking bullshit numbers - Added detailed duration info

MichalLustykNeti:
image

### DIFF
--- a/src/api/nodeRpcCall.js
+++ b/src/api/nodeRpcCall.js
@@ -1,6 +1,8 @@
 import { web3FromAddress } from '@polkadot/extension-dapp';
 import pako from 'pako';
-import { hexToU8a, u8aToHex } from '@polkadot/util';
+import {
+  hexToU8a, u8aToHex,
+} from '@polkadot/util';
 import { USER_ROLES, userRolesHelper } from '../utils/userRolesHelper';
 import { handleMyDispatchErrors } from '../utils/therapist';
 import { blockchainDataToFormObject } from '../utils/registryFormBuilder';
@@ -2020,6 +2022,16 @@ const createContract = async (data, parties, walletAddress) => {
   return submitExtrinsic(extrinsic, walletAddress, api);
 };
 
+const getStakingData = async (walletAddress) => {
+  const api = await getApi();
+  const [stakingInfo, sessionProgress] = await Promise.all([
+    api.derive.staking?.account(walletAddress),
+    api.derive.session.progress(),
+  ]);
+
+  return { stakingInfo, sessionProgress };
+};
+
 export {
   getBalanceByAddress,
   sendTransfer,
@@ -2127,4 +2139,5 @@ export {
   getSingleContract,
   createContract,
   getSignaturesForContracts,
+  getStakingData,
 };
--- a/src/components/Staking/StakeManagement/Unbonding/index.js
+++ b/src/components/Staking/StakeManagement/Unbonding/index.js
@@ -1,13 +1,35 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { useSelector } from 'react-redux';
+import { useSelector, useDispatch } from 'react-redux';
+import { BN_ZERO } from '@polkadot/util';
 import { formatDollars } from '../../../../utils/walletHelpers';
 import { blockchainSelectors, validatorSelectors } from '../../../../redux/selectors';
-import { eraToDays } from '../../../../utils/staking';
+import { blockTimeFormatted, stakingInfoToProgress } from '../../../../utils/staking';
+import { validatorActions } from '../../../../redux/actions';
 
-function UnbondingRow({ value, era }) {
-  const activeEra = useSelector(blockchainSelectors.activeEra);
-  const ready = activeEra.index.gte(era.unwrap());
+function UnbondingRow({ unlock, blocks }) {
+  return (
+    <li>
+      {formatDollars(unlock.value)}
+      {' '}
+      LLD
+      {' '}
+      {
+       `will unlock in ${blockTimeFormatted(blocks)}`
+      }
+    </li>
+  );
+}
+
+UnbondingRow.propTypes = {
+  // eslint-disable-next-line react/forbid-prop-types
+  unlock: PropTypes.object.isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  blocks: PropTypes.object.isRequired,
+};
+
+function UnbondingRowReady({ value }) {
+  if (!value || value.lte(BN_ZERO)) return null;
 
   return (
     <li>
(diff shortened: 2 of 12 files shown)


SzymonZiebaNeti (approved):

src/utils/staking.js
```diff
@@ -1 +1,29 @@
...
+    days,
+    hours,
+    minutes,
+  } = blockDuration(startBlock, endBlock);
+
+  const formattedDays = days <= 0 ? '' : `${days} Days`;
+  const formattedHrs = hours <= 0 ? '' : `${hours} Hrs`;
+  const formattedMins = minutes <= 0 ? '' : `${minutes} Mins`;
+
+  return [formattedDays, formattedHrs, formattedMins].join(' ');
```
kacperzuk-neti, line 28: If `endBlock - startBlock < 10` (a.k.a. days, hours & minutes are 0), this will return empty string. Instead, we should probably display something like `<1 min`.

kacperzuk-neti (changes requested):

kacperzuk-neti (approved):

src/components/Staking/StakeManagement/Unbonding/index.js
```diff
@@ -3,19 +3,21 @@ import PropTypes from 'prop-types';
...
+  const unlockBlock = era.toNumber() * 6 * 60 * 10;
 
   return (
     <li>
       {formatDollars(value)}
       {' '}
       LLD
       {' '}
-      {ready ? 'ready to withdraw' : `will unlock on ${eraToDays(era)} Days`}
+      {ready ? 'ready to withdraw' : `will unlock on ${blockDurationFormatted(blockNumber, unlockBlock)}`}
```
DorianSternVukotic, line 20: in instead of on

DorianSternVukotic (changes requested):

DorianSternVukotic:
Numbers still bullshit in our FE but ok in polkadot browser

Tested with FE connected on testnet

src/api/nodeRpcCall.js
```diff
@@ -2009,6 +2009,41 @@ const getIsUserJudges = async (walletAddress) => {
...
+          // otherwise we would enter into negative values (which is "accurate" since we are
+          // overdue, but confusing since it implied it needed to be done already).
+          //
+          // This does mean that in cases of era stalls we would have an jiggling time, i.e.
+          // would be down and then when a session completes, would be higher again, just to
+          // repeat the cycle again
+          //
+          // See https://github.com/polkadot-js/apps/issues/9397#issuecomment-1532465939
+          isStalled
+            ? progress.eraProgress.mod(progress.eraLength)
```
kacperzuk-neti, line 2041: This is directly copied from polkadot-js/apps. We need to check license compatibility and add proper licence/copyright notice.

src/utils/staking.js
```diff
@@ -1 +1,23 @@
+import {
+  extractTime, bnMin, BN_MAX_INTEGER, BN,
+} from '@polkadot/util';
+
 export const eraToDays = (era) => Math.round(era.toNumber() / 4);
```
kacperzuk-neti, line 5: `eraToDays` is unused now.

kacperzuk-neti (changes requested):

src/utils/staking.js
MichalLustykNeti: `eraToDays` is used in `UnbondModal` to make bondingDuration a time in days

kacperzuk-neti (approved):

DorianSternVukotic:
Says its ready to withdraw, but there is no withdraw unlocked button

According to polkadotjs, this is because its not actually ready to withdraw, so the numbers are still bs

src/components/Staking/StakeManagement/Unbonding/index.js
```diff
@@ -1,43 +1,91 @@
...
+        {stakingData && stakingData.map(({
+          unlock,
+          eras,
+          blocks,
+        }) => (
+          <UnbondingRow
+            key={eras.toString()}
+            {...{ unlock, eras, blocks }}
+          />
+        ))}
```
kacperzuk-neti, line 82: Why refactor like this? It's very messy and unreadable. Even to author, as it introduces a bug (<UnbondingRow> is shown always, even for these that are ready, resulting in duplicated rows).

src/components/Staking/StakeManagement/Unbonding/index.js
```diff
@@ -1,43 +1,91 @@
...
+  ) return null;
 
   return (
     <>
       Currently unstaking:
       <ul>
-        {info.unlocking.map(({ value, era }) => <UnbondingRow key={era.toString()} {...{ value, era }} />)}
+        {
+        info.unlocking.map(({ era, value }) => (
+          activeEra.index.gte(era)
```
kacperzuk-neti, line 69:
```suggestion
          activeEra.index.gte(era.unwrap())
```

`activeEra.index` is just a number (wrapped in BN), but `era` is `Compact<u32>`. Era needs to be unwrapped, otherwise this always returns true (and we get duplicated rows).

kacperzuk-neti (changes requested):

src/components/Staking/StakeManagement/Unbonding/index.js
kacperzuk-neti, line 82: This refactor is bad for readability and increases chances of bugs. I'd recommend leaving this in dedicated component that does the distinction.

1. `(args) => (condition && result)` is barely readable even in short cases, here it's totally unreadable
2. We now have two places where we check if unstaking is ready: filtering in nodeRpcCall and here (`activeEra.index...`). Easy to get inconsistencies.

kacperzuk-neti (changes requested):

src/components/Staking/StakeManagement/Unbonding/index.js
```diff
@@ -1,43 +1,85 @@
...
+    dispatch(validatorActions.getStakingData.call());
+  }, [dispatch, blockNumber, info]);
 
+  if (stakingData.length === 0 && (!stakingInfo?.redeemable || stakingInfo?.redeemable.lte(BN_ZERO))) return null;
   return (
     <>
       Currently unstaking:
       <ul>
-        {info.unlocking.map(({ value, era }) => <UnbondingRow key={era.toString()} {...{ value, era }} />)}
+        <UnbondingRowReady value={stakingInfo?.redeemable} />
```
kacperzuk-neti, line 66: Please update the `WithdrawUnbondedButton` to also rely on `stakingInfo.redeemable`. Right now it does era checks manually, which may be inconsistent.

kacperzuk-neti (changes requested):

kacperzuk-neti (approved):
Didnt test, but code looks fine

PR #271, opened 2024-04-16
BLOCKCHAIN-427 FE make companies its own menu entry, instead of being under registries

MichalLustykNeti:


.env.dist
```diff
@@ -29,6 +29,7 @@ REACT_APP_IS_LEGISLATION_DISCOURAGED=true
 REACT_APP_IS_OFFICES_DISCOURAGED=true
 REACT_APP_IS_REGISTRIES_DISCOURAGED=true
 REACT_APP_IS_CONGRESS_DISCOURAGED=true
+REACT_APP_IS_COMPANIES_DISCOURAGED=true
```
DorianSternVukotic, line 32: set to false

DorianSternVukotic (changes requested):

DorianSternVukotic:
Additional requirements.

Remove the 'encrypt field' checkboxes. Keep it as data, but the UI should not show it. False or null as default

Change 'append...' buttons text to 'add...' text

SzymonZiebaNeti (approved):

PR #274, opened 2024-04-22
Fix payouts when some validators are no longer stashes

kacperzuk-neti:
The issue happens when validator removes the stash completely by unbonding & withdrawing everything. I've recreated it locally and checked with Polkadot.js Apps and the same thing happens there - they show up as pending rewards, but can't be paid out:
Screenshot from 2024-04-22 13-36-37

On it's own it wouldn't be a big issue, but we're batching payouts for multiple validators, so it actually blocks user from paying our anything for 84 eras (as that's when the broken one will expire). Polkadot.js Apps does allow to do payouts only on specific validators, so they're not affected as much as we are.

Solution: just remove these broken validators from any calculations & payouts.

### DIFF
--- a/src/api/nodeRpcCall.js
+++ b/src/api/nodeRpcCall.js
@@ -1300,15 +1300,33 @@ const getStakersRewards = async (accounts) => {
   // allRewards may include rewards from validators that no longer have a stash.
   // Such rewards are unclaimable and essentially lost either way, so let's just
   // filter them out.
+
+  // First find all validators that our stakers got rewards from
   const uniqValidatorsSet = new Set(allRewards.flatten().map(({ validators }) => Object.keys(validators)).flatten());
   const uniqValidators = Array.from(uniqValidatorsSet);
-  const ledgers = await api.query.staking.ledger.multi(uniqValidators);
-  const brokenValidators = ledgers.reduce((broken, v, i) => {
+
+  // Validators are stashes, convert to controllers
+  const controllers = await api.query.staking.bonded.multi(uniqValidators);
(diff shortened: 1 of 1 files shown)


kacperzuk-neti:
Test log:
image
image
With this state, Bob unbonds and withdraws everything. It disappears from stashes, but is still on payouts, which don't work:
image
image

Our FE shows some pending rewards:
image
Which can be paid out
image
And on polkadot.js apps we can verify that we did pay out the part from good validator (Alice Stash disappeared from the list):
image

MichalLustykNeti (approved):

SzymonZiebaNeti (approved):

DorianSternVukotic:
I think its bugged.

Account 5G3uZjEpvNAQ6U2eUjnMb66B8g6d8wyB68x6CfkRPNcno8eR has payouts in polka but not in our site

kacperzuk-neti:
> I think its bugged.
> 
> Account 5G3uZjEpvNAQ6U2eUjnMb66B8g6d8wyB68x6CfkRPNcno8eR has payouts in polka but not in our site

Yup, there was a bug - it filtered out rewards from validators that have split stash & controller. In practice this means it filtered out our genesis validators, as citizen validators use same account for stash & controller AFAIK. 
Either way, fixed in latest commit. Tested both by spoofing `5G3uZjEpvNAQ6U2eUjnMb66B8g6d8wyB68x6CfkRPNcno8eR` and by recreating complete thing on dev like in https://github.com/liberland/liberland_frontend/pull/274#issuecomment-2069271636

PR #277, opened 2024-04-23
BLOCKCHAIN-419 FE - transaction history - amount show +0

MichalLustykNeti:


src/utils/walletHelpers.js
```diff
@@ -51,7 +51,11 @@ export const formatTransaction = (value_raw, bigSymbol, smallSymbol, decimals, c
   const absIntvalue = value.abs();
 
   if (_parse(absIntvalue.toString(), decimals).gt(BN_ONE)) {
-    const formatValue = _format(absIntvalue, config.isAsset ? parseInt(decimals) : decimals);
+    const formatValue = ethers.utils.formatUnits(
```
kacperzuk-neti, line 54: Instead of switching to ethers (which I hope we can get rid of someday), lets keep _format, just use the `withAll` parameter.
Right now we're not passing it to `formatBalance`, but we add it in the DEX PR: https://github.com/liberland/liberland_frontend/blob/ebff469717d674cd78ae28974cbe908970ef1101/src/utils/walletHelpers.js#L20-L30
Feel free to copy it from DEX PR verbatim.

kacperzuk-neti: Docs: https://github.com/polkadot-js/common/blob/master/packages/util/src/format/formatBalance.ts#L32-L35

kacperzuk-neti (changes requested):

kacperzuk-neti (approved):

SzymonZiebaNeti (approved):

PR #280, opened 2024-04-25
BLOCKCHAIN-422 FE - Add HASHI SORA bridge transactions to transaction history

MichalLustykNeti:
image
image

### DIFF
--- a/src/api/explorer.js
+++ b/src/api/explorer.js
@@ -11,7 +11,10 @@ const historyTransferQuery = `
     $orderByAssetTransfers: [AssetTransfersOrderBy!],
     $filterAssetTransfers: AssetTransferFilter,
     $orderByStakings: [StakingsOrderBy!],
-    $filterStakings: StakingFilter
+    $filterStakings: StakingFilter,
+    $filterSoraMinted: SoraMintedFilter,
+    $filterSoraBurned: SoraBurnedFilter,
+  
   ) {
     transfers(orderBy: $orderByTransfers, filter: $filterTransfers) {
       nodes {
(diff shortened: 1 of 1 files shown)


src/api/explorer.js
```diff
@@ -79,6 +80,29 @@ const historyTransferQuery = `
         }
       }
     }
+
+    events(filter: {
```
kacperzuk-neti, line 84: Damn, my bad, the query in ticket was wrong. `events` will be inefficient, as `data.contains` filter is not indexed in DB. Use `soraMinteds` and `soraBurneds` like this: https://github.com/liberland/chain_explorer/pull/17
Sorry!

kacperzuk-neti (changes requested):

src/api/explorer.js
```diff
@@ -153,27 +207,78 @@ export const getHistoryTransfers = async (substrate_address) => {
...
+  const soraMintedParsed = soraMinted.map(({
+    asset, block, recipientId, value,
+  }) => {
+    const parsedAsset = JSON.parse(asset);
+    const isLLD = Object.keys(parsedAsset)[0] === 'lld' && parsedAsset.lld === null;
+
+    return {
+      asset: isLLD ? 'LLD' : assetsData[parsedAsset.asset].metadata.symbol,
+      decimals: isLLD ? null : assetsData[parsedAsset.asset].metadata.decimals,
+      fromId: soraBridgeName,
```
kacperzuk-neti, line 228: Technically we have data on the exact Sora Network account that is the sender (for mints) and the recipient (for burns). I think just putting `Sora Bridge` is fine, just leaving this as comment for Dorian in case he wants it different.

kacperzuk-neti (approved):

SzymonZiebaNeti (approved):

PR #282, opened 2024-04-30
fix mobile menu overflow, fix mobile select wallet

SzymonZiebaNeti:


src/components/Home/ChangeWallet/styles.module.scss
```diff
@@ -11,11 +11,35 @@
 
   .select{
     width: 100%;
+    height: 50px;
     border-radius: 6px;
     padding: 11px 13px;
+    appearance: none; 
+    border: 1px solid $secondary3;
```
DorianSternVukotic, line 18: wasnt a requirement but looks nicer now, thanks

src/components/Home/ChangeWallet/styles.module.scss
```diff
@@ -11,11 +11,35 @@
...
+
+    option {
+      margin-top: 10px;
+      background-color: $bg4; 
+    }
+  }
+
+  &::after {
+    content: '\2039'; 
+    position: absolute;
```
DorianSternVukotic, line 39: why this after ?

SzymonZiebaNeti: So there is a problem with styling arrow in select, so we can make custom arrow on wrapper, but i found a new solution to this, and it's more like the default arrow on select.

PR #288, opened 2024-05-20
BLOCKCHAIN-447 - FE dont onboard if already onboarded

SzymonZiebaNeti:


src/components/GuidedSetup/index.js
```diff
@@ -88,6 +92,16 @@ function GuidedSetup({ children }) {
     }
   }, [identityData]);
 
+  useEffect(() => {
+    if (!isIdentityEmpty) {
```
DorianSternVukotic, line 96: Why use this as useEffect instead of just putting it next to where setIsIdentityEmpty is used?

DorianSternVukotic (changes requested):

PR #293, opened 2024-06-05
Bugfix legislation content view

kacperzuk-neti:


src/components/Legislation/LegislationView/index.js
```diff
@@ -421,7 +422,7 @@ function LegislationView() {
     dispatch(legislationActions.getCitizenCount.call());
     // required by RepealLegislationButton && CitizenProposeRepealLegislationButton
     dispatch(congressActions.getMembers.call());
-  }, [dispatch, tier, legislationActions]);
+  }, [dispatch, tier]);
```
kacperzuk-neti, line 425: This is not directly related to the PR, but pre-commit hook was preventing me from commiting this due to:
```
/home/kacper/liberland/liberland_frontend/src/components/Legislation/LegislationView/index.js
  425:6  warning  React Hook useEffect has an unnecessary dependency: 'legislationActions'. Either exclude it or remove the dependency array. Outer scope values like 'legislationActions' aren't valid dependencies because mutating them doesn't re-render the component  react-hooks/exhaustive-deps
```

PR #297, opened 2024-06-27
fix assertion error, not show remove liquidity when user don't have l…

SzymonZiebaNeti:
…iquidity

### DIFF
--- a/src/components/Wallet/Exchange/ExchangeShowMore/index.js
+++ b/src/components/Wallet/Exchange/ExchangeShowMore/index.js
@@ -80,7 +80,7 @@ function ExchangeShowMore({
 
         </div>
         <div className={styles.liquidityButtons}>
-          {!isReservedDataEmpty
+          {!isReservedDataEmpty && (lpTokensBalance !== 0 || !liquidity)
             && (
             <Button small green onClick={handleModalLiquidityRemove}>
               Remove Liquidity
--- a/src/components/Wallet/Exchange/ExchangeShowMore/index.js
+++ b/src/components/Wallet/Exchange/ExchangeShowMore/index.js
@@ -41,17 +41,14 @@ function ExchangeShowMore({
       </div>
       <div className={styles.liquidity}>
         <div className={styles.text}>
-          {lpTokensBalance !== 0 || !liquidity
+          {!lpTokensBalance.isZero() || !liquidity
             ? (
               <>
                 <span>
                   Your liquidity:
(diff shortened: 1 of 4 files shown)


src/components/Wallet/Exchange/ExchangeShowMore/index.js
```diff
@@ -49,7 +49,7 @@ function ExchangeShowMore({
                   {' '}
                   {new BN(lpTokensBalance).mul(BN_HUNDRED).div(new BN(liquidity)).toString()}
                   % (
-                  {lpTokensBalance}
+                  {lpTokensBalance.toString()}
```
kacperzuk-neti, line 52: Hmmm... lpTokensBalance is a BN here, right? I'd worry that `.toString()` may return hex for some values here. Maybe use formatAssets with decimals=0 instead?
Also, what values are seen here in practice? I'd worry that these are gigantic number, while at the same time it doesn't really make sense to put arbitrary decimals on it... Is it needed at all? Just keeping percent seems enough.

src/components/Wallet/Exchange/ExchangeShowMore/index.js
```diff
@@ -49,7 +49,7 @@ function ExchangeShowMore({
                   {' '}
                   {new BN(lpTokensBalance).mul(BN_HUNDRED).div(new BN(liquidity)).toString()}
```
kacperzuk-neti, line 50: `lpTokensBalance` already is a BN now?

src/components/Wallet/Exchange/ExchangeShowMore/index.js
```diff
@@ -80,7 +80,7 @@ function ExchangeShowMore({
 
         </div>
         <div className={styles.liquidityButtons}>
-          {!isReservedDataEmpty
+          {!isReservedDataEmpty && (lpTokensBalance !== 0 || !liquidity)
```
kacperzuk-neti, line 83: `lpTokensBalance !== 0` probably won't work - `lpTokensBalance` is now an object (BN probably).

kacperzuk-neti (changes requested):

src/components/Wallet/Exchange/ExchangeShowMore/index.js
DorianSternVukotic: yeah just keep percent then

kacperzuk-neti (approved):

PR #305, opened 2024-07-23
BLOCKCHAIN-459 FE Congress budget proposal

SzymonZiebaNeti:


src/components/Congress/index.js
```diff
@@ -11,8 +12,26 @@ import Wallet from './Wallet';
...
+  useEffect(() => {
+    dispatch(officesActions.getPalletIds.call());
+  }, [dispatch]);
+
+  useEffect(() => {
+    dispatch(congressActions.congressGetWallet.call());
+  }, [dispatch, pallets]);
+
+  useEffect(() => {
+    dispatch(congressActions.congressGetAdditionalAssets.call());
```
DorianSternVukotic, line 32: i really dont like this pattern of using useEffects to make a chained of calls one after the other.

Is this really an accepted pattern @kacperzuk-neti @SzymonZiebaNeti ?

src/components/WalletCongresSenate/RemarkForm/index.js
```diff
@@ -0,0 +1,128 @@
...
+      />
+
+      <div className={styles.title}>Project</div>
+      <TextInput
+        register={register}
+        name={projectName}
+        errorTitle="Project"
+        placeholder="Project"
+        required
+        maxLength={{ value: 256, message: 'Max Project length is 256 chars' }}
```
DorianSternVukotic, line 69: I thought max remark length in total is 256, not individual items?

src/components/Congress/index.js
```diff
@@ -11,8 +12,26 @@ import Wallet from './Wallet';
...
+import { congressActions, officesActions } from '../../redux/actions';
+import { congressSelectors, officesSelectors } from '../../redux/selectors';
 
 function Congress() {
+  const dispatch = useDispatch();
+  const pallets = useSelector(officesSelectors.selectorPallets);
+  const congresAccountAddress = useSelector(congressSelectors.walletAddress);
+
+  useEffect(() => {
+    dispatch(officesActions.getPalletIds.call());
```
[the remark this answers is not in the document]
kacperzuk-neti: We need pallet its to know the congress account address (it's derived from congress custom account pallet id).

src/components/Congress/index.js
kacperzuk-neti: I'm not a Redux Sagas expert, but I don't like it either. it should be combined in a single action (which, if needed, can dispatch actions it depends on its of in the saga if needed).

src/components/WalletCongresSenate/RemarkForm/index.js
kacperzuk-neti: Correct, total remark length must be <= 256

DorianSternVukotic:
- Remove verification for amounts - congress can propose budgets for higher amounts than it currently has
- LLD budget proposal makes a assets transfer call, this is wrong ?
- close modal after submitting
- add 'final destination' field. Sometimes congress will send USDT to an escrow account on LL chain with final destination on ETH, for example

src/utils/councilHelper.js
```diff
@@ -0,0 +1,63 @@
...
+      remark: objectToHex(remark),
+    };
+  });
+  return resultArray;
+};
+
+export const OperationsType = {
+  LLD: 'LLD',
+  LLM: 'LLM',
+  POLTIPOOL_LLM: 'POLTIPOOL_LLM',
```
kacperzuk-neti, line 43:
```suggestion
  POLITIPOOL_LLM: 'POLITIPOOL_LLM',
```
Typo.

src/components/Modals/ProposeBudgetModal.js
```diff
@@ -0,0 +1,264 @@
...
+  const onSubmit = (data) => {
+    if (!isValid) return;
+    const bugetProposalItems = extractItemsFromObject(data, spendings[0].optionsInput);
+    dispatch(congressActions.congressBudgetPropose.call({ bugetProposalItems, executionBlock }));
+  };
+
+  const optionsInputDefault = useMemo(() => allBalance.map((item) => {
+    const { metadata, index, balance } = item;
+    const { symbol, name, decimals } = metadata;
+    const operationType = getOperationType(symbol);
```
kacperzuk-neti, line 79: We can't abuse symbol for choosing transfer type here.
1. Asset symbols are chosen by people, anyone can create asset with a symbol "LLD" which would break this system here
2. It's just error-prone and I'd guess this may be related to error reported by @DorianSternVukotic (where LLD transfer used asset.transfer onchain)

DorianSternVukotic:
Remove validation of final destination as it can be literally anything

"bitcoin adress XYZ"
"ETH address XYZ"
"Money will be used to buy gold and hand delivered via clowns to physical adress XYZ..."

PR #307, opened 2024-07-26
Blockchain 464 pretty print congress motions for lld assets spendings

SzymonZiebaNeti:


DorianSternVukotic:
Please rebase into proper origin (the congress budgets)

Please add some space between each proposal (right now its a block of text, we just want one <br> in motions with multiple proposals)

PR #311, opened 2024-08-14
BLOCKCHAIN-483 FE Congress close motions that have too many nay votes

SzymonZiebaNeti:


DorianSternVukotic:
50% nay votes reached, no button showing

src/components/WalletCongresSenate/Motion/index.js
```diff
@@ -25,6 +26,8 @@ export default function Motion({
 
   const isClosable = voting.ayes.length >= threshold;
 
+  const isClosableNaye = voting.nays.length >= 50;
```
kacperzuk-neti, line 29: It should be something like `voting.nays.length > number_of_congressmen - threshold`.

SzymonZiebaNeti: yes i just take `has more than 50 nay votes` from task, too literally

DorianSternVukotic: percent xD

DorianSternVukotic (changes requested):
0 nays, close motion button present, it shouldnt be like this  ?

src/components/WalletCongresSenate/Motion/index.js
```diff
@@ -58,7 +61,7 @@ export default function Motion({
               <b>
                 {voting.nays.length}
                 /
-                {threshold}
+                {(membersCount / 2).toFixed()}
```
kacperzuk-neti, line 64: Doesn't make sense. Why /2?

src/components/WalletCongresSenate/Motion/index.js
```diff
@@ -25,6 +26,8 @@ export default function Motion({
 
   const isClosable = voting.ayes.length >= threshold;
 
+  const isClosableNaye = voting.nays.length >= membersCount - threshold;
```
kacperzuk-neti, line 29: This should be a hard inequality (`>`).
See my previous comment: https://github.com/liberland/liberland_frontend/pull/311#discussion_r1726898829

Seems like there's still no understanding of how this works, so let me explain.

Motion can be "closed & executed", when there's enough `ayes` votes to meet the threshold.
Motion can be "closed", when there's enough `nays` that we already know that we can't get enough `ayes` anymore to meet the threshold.

So a few samples:
* 3 members, threshold=2. So 1 nay is OK, but if we get 2 nays then there's only 1 more member left to 
potentially cast aye, so we should show the `close` button
* 4 members, threshold=3. So same as before - 1 nay is ok, but 2 nays leaves us with only 2 more potential ayes, which isn't enough for the threshold
* 10 members, threshold=6. Up to 4 nays is OK, as it still leaves 6 members to potentially cast aye. 5 nays is when we know that this will never pass and we can close.

src/components/WalletCongresSenate/Motion/index.js
kacperzuk-neti: If anything this should be `membersCount-threshold+1`, but TBH I expect this to confuse the hell out of people.

kacperzuk-neti (changes requested):

PR #314, opened 2024-08-22
Fix remark block submit

SzymonZiebaNeti:


src/api/nodeRpcCall.js
```diff
@@ -14,6 +14,7 @@ import { parseDollars, parseMerits } from '../utils/walletHelpers';
 import { getMetadataCache, setMetadataCache } from '../utils/nodeRpcCall';
 import { addReturns, calcInflation, getBaseInfo } from '../utils/staking';
 import identityJudgementEnums from '../constants/identityJudgementEnums';
+// eslint-disable-next-line import/no-cycle
```
DorianSternVukotic, line 17: Do we care about cycling ?

src/components/Proposal/index.js
```diff
@@ -338,18 +338,32 @@ function TransferAsset({ proposal }) {
...
+        finalDestination: item.finalDestination,
+      });
+    });
+  }, [bytes, proposal]);
+
+  if (!data) {
     return (
       <div>
-        {jsonString}
+        {JSON.stringify(proposal.toHuman(), null, 2)}
```
DorianSternVukotic, line 362: didnt we say not to use toHuman?

DorianSternVukotic:
When doing budgets amount in USD always shows as 0, single spendings are fine

DorianSternVukotic (changes requested):

PR #315, opened 2024-08-27
BLOCKCHAIN-425 - FE enable usage without logging in

SzymonZiebaNeti:
We want to remove singin page, so in env file, we need to delete signin path from env (in .env.dist)
We change the redirect page on frontend, so we need also to change it in sso database

ll_auth_development => clients => callbackUrl => {http://localhost:8080/singIn} change to {http://localhost:8080}, so if it's looks same on test or production, it's needed to change it also there.

image


src/components/CheckExtenstionWalletProvider/index.js
```diff
@@ -0,0 +1,78 @@
...
+  useEffect(() => {
+    (async () => {
+      setIsBrave(!!(navigator.brave && (await navigator.brave.isBrave())));
+    })();
+  }, [setIsBrave]);
+
+  return isBrave;
+};
+
+export function CheckExtensionWalletProvider({ children }) {
```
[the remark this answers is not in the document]
SzymonZiebaNeti: Copied, i make 2 separated components for check extensions / wallets, and other for onboarding

DorianSternVukotic:
Mobile navigation doesnt work

src/App.js
```diff
@@ -21,6 +21,7 @@ import { userSelectors } from './redux/selectors';
 import { authActions } from './redux/actions';
 import GuidedSetup from './components/GuidedSetup';
 import { CheckExtensionWalletProvider } from './components/CheckExtenstionWalletProvider';
+import SignIn from './components/AuthComponents/SigIn';
```
DorianSternVukotic, line 24: typo, should be SignIn

DorianSternVukotic (changes requested):

src/redux/selectors/blockchainSelectors.js
```diff
@@ -12,12 +12,12 @@ const allWalletsSelector = createSelector(
 );
 const userWalletAddressSelector = createSelector(
   blockchainReducer,
-  (reducer) => reducer.userWalletAddress,
+  (reducer) => reducer?.userWalletAddress,
```
kacperzuk-neti, line 15: `reducer` should never be empty. It's set here https://github.com/liberland/liberland_frontend/blob/BLOCKCHAIN-425-fe-enable-usage-without-logging-in/src/redux/reducers/index.js#L27 and it has an initial state: https://github.com/liberland/liberland_frontend/blob/BLOCKCHAIN-425-fe-enable-usage-without-logging-in/src/redux/reducers/blockchain.js#L5

Did you experience it being empty in practice? If yes, this suggest we have some really nasty bug somewhere that overwrites whole blockchain state and we should address it instead of defaulting with `?`. If not, then `?` isn't needed here.

src/redux/selectors/blockchainSelectors.js
```diff
@@ -12,12 +12,12 @@ const allWalletsSelector = createSelector(
...
 );
 const isUserWalletAddressSameAsUserAdress = createSelector(
   (state) => state,
-  (reducer) => reducer.user.user.blockchainAddress
-      === reducer.blockchain.userWalletAddress,
+  (reducer) => reducer?.user?.user?.blockchainAddress
+      === reducer?.blockchain?.userWalletAddress,
```
kacperzuk-neti, line 20: Ditto, `reducer`, `reducer.blockchain` and `reducer.user` should never be empty here. `reducer.user.user` is the only one that can be null in practice.
If that's not the case, there's a really nasty bug somewhere.

src/redux/selectors/walletSelectors.js
```diff
@@ -44,7 +44,7 @@ const selectorTotalBalance = createSelector(
 const selectorTotalLLM = createSelector(
   selectorBalances,
   (reducer) => (
-    reducer.meritsTotalAmount.amount
+    reducer?.meritsTotalAmount?.amount
```
kacperzuk-neti, line 47: Ditto, `reducer` and `reducer.meritsTotalAmount` should never be empty, if they are then we have some serious issue somewhere.

src/redux/selectors/blockchainSelectors.js
SzymonZiebaNeti: yea, nice catch, it should be only for blockChainAddress on user, when it's not logged in.

PR #317, opened 2024-08-28
BLOCKCHAIN-503 Show identity name on motions, congress/senate  / fix …

SzymonZiebaNeti:
…showing invalid address in search identity forms


DorianSternVukotic (changes requested):
Bug with resolving (i assume undefined passed variable) to random address is back

src/components/Congress/Motions/index.js
```diff
@@ -15,13 +15,18 @@ export default function Motions() {
     dispatch(congressActions.getMotions.call());
   }, [dispatch]);
 
-  if (!motions || motions.length < 1) {
+  useEffect(() => {
+    if (!motions?.names) return;
+    dispatch(identityActions.getIdentityMotions.call(motions.names));
```
DorianSternVukotic, line 20: why do it as useeffect instead of doing it in the call or as a param when fetching the data, especially given this appears in multiple places?

DorianSternVukotic (changes requested):

DorianSternVukotic:
conflicts

PR #324, opened 2024-09-30
Fix/show display legal

SzymonZiebaNeti:
check after or compare to https://github.com/liberland/liberland_frontend/pull/323


SzymonZiebaNeti:
When i made it i didn't know that you store legal name in additional info, so here is just info direct from legal property in info.

In next task i make whole logic, which one taking care of any property in additional data. So will also work for legal additional

https://github.com/liberland/liberland_frontend/pull/326

PR #325, opened 2024-10-02
BLOCKCHAIN-511 FE onboarding improvements

SzymonZiebaNeti:


DorianSternVukotic (changes requested):
Dont know if caused by this PR or undiscovered bug but i get white screen after changing my adress in database to nonexistent one, clearing all memory and cache, logging in, skipping onboarding LLD and identity, and when it asks to change registered adress, i click connect and get white screen of death

Console complains
18             VEC: Unable to decode on index 0 createType(Lookup0):: Decoding 5GjYePC6HKJGGnEzEZzgfdgfdSvimy6uctuMat4Kr2tjACtKyY9nhT: Invalid decoded address length

I guess its because the address is bullshit. But users are able to set centralized address to bullshit, and this step is after clicking on changing that address to existing one

Anyway needs fixing.

Also, if user logs in, has centralized address, the address questions should come BEFORE claim LLD and update identity.

PR #346, opened 2024-11-06
Features/add eth direct staking

jindra12:
Add direct ETH staking form that displays how much LP the user gets


package.json
```diff
@@ -40,6 +41,7 @@
     "postcss-normalize": "^10.0.1",
     "postcss-preset-env": "^7.4.3",
     "process": "^0.11.10",
+    "rc-slider": "^11.1.7",
```
DorianSternVukotic, line 44: we already use a slider in LLDEX when removing liquidity

DorianSternVukotic (changes requested):

PR #347, opened 2024-11-06
BLOCKCHAIN-516 FE enable usage without any wallet

SzymonZiebaNeti:


src/redux/sagas/base.js
```diff
@@ -1,17 +1,22 @@
...
+import { blockchainActions, routeActions } from '../actions';
 
 function errorHandler(onFailure, worker) {
   function* errorHandledWorker(action) {
     try {
       yield* worker(action);
     } catch (e) {
       // eslint-disable-next-line no-console
-      console.error(e);
+      const isWallet = !e.stack.includes('Error: web3FromAddress: Unable to find injected');
```
jindra12, line 12: This is very defensive and should be handled by reviewing the codebase and ensuring this doesn't happen to users. At the very least, this should be an "instanceof" check for error type if possible.

jindra12: For instance, what if we upgrade polkadot library and  they change the error messages? This would stop working then.

jindra12: If this is about accessing routes without an appropriate browser/wallet combination, I think react-router-dom library already has a way of validating a route, maybe try that?

src/components/CheckExtenstionWalletProvider/index.js
```diff
@@ -59,15 +58,7 @@ export function CheckExtensionWalletProvider({ children }) {
...
-    return (
-      <GuidedSetupWrapper>
-        <NoWalletsDetectedInBrowser />
-      </GuidedSetupWrapper>
-    );
-  }
-
-  if (!walletAddress) {
+  if (!walletAddress && (userBlockchainAdressStorage || wallets[0]?.address)) {
     dispatch(blockchainActions.setUserWallet.success(userBlockchainAdressStorage || wallets[0].address));
```
jindra12, line 62: Can you also fix this please? This should be an useEffect not a part of render

src/components/CheckExtenstionWalletProvider/index.js
```diff
@@ -41,6 +40,16 @@ export function CheckExtensionWalletProvider({ children }) {
     || wallets === null
     || isUnsupportedBrowser === null;
 
+  useEffect(() => {
+    if (!walletAddress && (userBlockchainAdressStorage || (wallets?.length > 0 && wallets[0]?.address))) {
+      dispatch(
+        blockchainActions.setUserWallet.success(
+          userBlockchainAdressStorage || wallets[0].address,
```
DorianSternVukotic, line 47: hmmm will have to see how this works with wallets that require you to connect again to the site

src/components/Home/ChangeRouter.js
```diff
@@ -0,0 +1,26 @@
...
+  const dispatch = useDispatch();
+
+  const history = useHistory();
+
+  useEffect(() => {
+    if (routeLink) {
+      setTimeout(() => {
+        history.push(routeLink);
+        dispatch(routeActions.changeRoute.call());
+      }, 2000);
```
DorianSternVukotic, line 18: why is this on timeout, what is this doing?

src/redux/sagas/blockchain.js
```diff
@@ -66,7 +66,7 @@ export function* subscribeWalletsSaga() {
       const wallets = await web3Accounts();
       emitter({ extensions, wallets });
     };
-    updateWallets();
+    setTimeout(updateWallets, 500);
```
DorianSternVukotic, line 69: why

DorianSternVukotic:
Rebase

White screen of death on clicking wallet - send lld without wallet
When signing transaction without wallet i get notification no wallet detected, you need to login and then it automatically redirects. Show a message to user, thats fine, but dont auto redirect, add a login and cancel button when signing without wallet

src/components/ErrorModal/index.js
```diff
@@ -21,14 +22,29 @@ function ErrorModal({ children }) {
...
+                  Ok
+                </Button>
+              )
+              : (
+                <div className={styles.buttons}>
+                  <Button
+                    secondary
+                    onClick={() => dispatch(blockchainActions.setErrorExistsAndUnacknowledgedByUser.call(false))}
+                  >
+                    Canel
```
DorianSternVukotic, line 41: canCel

DorianSternVukotic (changes requested):

PR #326, opened 2024-10-02
Blockchain 510 fe identity character limits

SzymonZiebaNeti:
Check after or compare to https://github.com/liberland/liberland_frontend/pull/324

PR #349, opened 2024-11-06
Blockchain 534 fe congress and senate motions voters

SzymonZiebaNeti:
Using logic from Blockchain-510 so compare to or check after - https://github.com/liberland/liberland_frontend/pull/326


DorianSternVukotic:
conflicts

PR #355, opened 2024-11-13
BLOCKCHAIN-533 FE scheduled congress spendings identities

SzymonZiebaNeti:


DorianSternVukotic (changes requested):
Ok so now if there is a scheduled congress spending, and senate starts motion to veto it, senate motions will have white screen of death. Probably missing identity provider

PR #370, opened 2024-12-22
Features/antd optimization

jindra12:


DorianSternVukotic:
Split webpack into multiple instead of one huge file
Import components conditionally, not everything at once
Split big layout component into multiple so that it rerenders less often.

PR #372, opened 2025-01-16
Features/fix minor errors and implement modern contracts

jindra12:


DorianSternVukotic:
1) increase max width of contract list 
2)  Contracts listFix alignment of contract info, title and text should be more or less aligned
3) Contracts list on big screens dont center it, align left
4) add spacing between individual contracts in list
5) on mobile reduce the gap between title and content
6) In detail view, ADD buttons to sign and copy on bottom, so they are both on top AND bottom
7) Implement party signature per design and fix its width so 3 ppl fit in one row. Use empty profile images for now

PR #373, opened 2025-01-18
Features/implement company page

jindra12:


jindra12:
Comment from @DorianSternVukotic

1) Regarding company link if empty, pls include the first letter of name as profile like you do for wallet addresses
2) company list company item description align center height wise, so if there is one line description its aligned with company name
3) website links dont work in any placce - goes to localhost/website
4) looking at your own companies, then clicking show info, theres a back to all companies button. But, i think it should just be a back button, taking you to your companies, or all companies depending on how you got there (with  default being all companies) 
5) company creation and edit move Add and delete buttons on the left instead of right. ppl mostly scroll with right thumb :P
6) Company mobile view does not correspond to design. but keep the website big button instead of just website link, just fix layout

DorianSternVukotic:
company list website button shows even if company doesnt have website - remove charterurl as website

company list mobile view needs bigger margins for description row, both top and bottom.

Register a new company button is so outt of the way. Can we also put it on the bottom of my companies list?

Company create, align the buttons for add/delete if not too difficult

PR #376, opened 2025-01-22
Features/add proper staking page

jindra12:


DorianSternVukotic (changes requested):
Also  change SOL staking LP link to https://raydium.io/clmm/create-position/?pool_id=3gYXPAYp8xi7wrzC9hpa89d2pQ8jBqT1UUL5wehvz3LY

staker mode nominated field - remove "nominated" string, keep just the checkbox and make the checkbox bigger

Selecting validators to nominate and then navigating away prompts a "you have unsaved changes" box with 2 options, cancel and ok. Rename them to discard and apply. Also change that whole component to a normal alert because in current implementation it doesnt work in mobile

Nominator mode move update nominations button to the right

Add a component on top: Current mode: validator/nominator with switch mode button there

Unstaking component - add info about the sum total of LLD currently in the process of unstaking so it will say how many are ready to withdraw and how many are currently unbonding

Nominators list with all the validators should be renamed to validators

Newly renamed validators list should not appear in validator mode

Bring back the reward graph for validators, why was it deleted? Optionally implement redesign design

DorianSternVukotic:
Checkbox  for selecting validators needs to be bigger centered, border bigger, and less gray - More contrast! Users are retarded, illiterate, and blind. It needs to be easy to see where they should click

Did unstaking details get deleted? Something is wrong with the logic now, it only shows when the next batch of LLDs will be ready, it hides the info about other batches of unstaking being ready

In nominator mode, the list of validators is named "nominators" Rename it to validators

Alert for leaving page with unsaved changes is wrong. Just make it a normal browser alert?

Switching to validator, staking rewards section is there but its just... empty

jindra12:
> Checkbox for selecting validators needs to be bigger centered, border bigger, and less gray - More contrast! Users are retarded, illiterate, and blind. It needs to be easy to see where they should click
> 
> Did unstaking details get deleted? Something is wrong with the logic now, it only shows when the next batch of LLDs will be ready, it hides the info about other batches of unstaking being ready
> 
> In nominator mode, the list of validators is named "nominators" Rename it to validators
> 
> Alert for leaving page with unsaved changes is wrong. Just make it a normal browser alert?
> 
> Switching to validator, staking rewards section is there but its just... empty
> 
> Validator creation menu - autocomplete commission to 20
> 
> when you unstake the UI autocompletes the max unstaking amount but keeps " , " so like 1,025.555. Remove the "," as its invalid char for number. Same for adding stake

Fixed everything but staking rewards section, since there's really nothing to show. Instead, I added an alert to tell you there are no rewards detected lol

src/components/Staking/Nominator/Nominated/styles.module.scss
```diff
@@ -1,3 +1,4 @@
 .checkbox {
     zoom: 1.5;
+    margin-left: 17px;
```
DorianSternVukotic, line 3: is that really the best way to center it lol?

DorianSternVukotic:
There should be 3 buttons

2 main ones
Start staking as nominator - asks you to stake LLD
Start validator server - create validator modal
Learn more  - leads to https://docs.liberland.org/blockchain/for-validators-nominators-and-stakers/staking

Also pls remove start validator button from the validator interface - you are already a validator at that point

src/components/Staking/Nominator/Nominated/styles.module.scss
jindra12: Yes, the cell itself is too big and it looks out of place since table headings aren't centered (that would look uglier)

src/components/Staking/Validator/CreateValidatorButton/index.js
```diff
@@ -10,11 +10,7 @@ export default function CreateValidatorButton() {
   const handleModalOpen = () => setIsModalOpen(!isModalOpen);
 
   if (info.stash) {
```
DorianSternVukotic, line 12: the parent function is dead?

PR #378, opened 2025-01-29
Features/hammer out last changes before big merge

jindra12:


DorianSternVukotic:
Add token icons to the modal headers of exchanges please

Also, the switch to registered wallet is supposed to be a button (yellow!) on the top of the navigation menu, not hidden user profile menu

PR #384, opened 2025-02-05
Blockchain 570 fe dex simpler design

SzymonZiebaNeti:
This branch used modal refactor branch.
https://github.com/liberland/liberland_frontend/pull/377

### DIFF
--- a/src/components/Loader/index.js
+++ b/src/components/Loader/index.js
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import Flex from 'antd/es/flex';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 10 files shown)


DorianSternVukotic (changes requested):
Swap TEUR for LLD on testnet fails with self.words is null

Usually sell is on top and buy is on bottom, so switch them please

Switching tokens via the button behaves weird, just copy uniswap logic. So for example, if you have sell 1 ETH buy 3000 USDT, clicking switch should turn into sell 3000 USDT buy 1 ETH, not sell 1 USDT buy 0.003 ETH or whatever

DorianSternVukotic:
On stock exchange on testnet, there are 2 different bugs, both happen when swapping token places after inputing value then trying to execute swap. Please test various combinations of entering values and swapping in stock exchange and fix the bugs.

PR #386, opened 2025-02-11
Features/fe edits batch

jindra12:


DorianSternVukotic (changes requested):
Switching wallet on mobile should close the menu. Right now an added regression is that switching pages on mobile doesnt close the menu.

Minting NFT doesnt work as pinata returns invalid request format 400 error and mint fails with nfts.UnknownItem: The given item ID is unknown.

Propose budget for senate is actually proposing budget for congress. Nevermind. Remove that requirement and remove that button from Senate

PR #396, opened 2025-02-19
Features/connect assets and stocks

jindra12:


DorianSternVukotic:
Dont truncate stuff in company creation forms

Company create/edit form whenever it asks for passport nmber change to "Liberland passport or ID number"

Company detail, web address buttons are wrong when you click on them.

When company and stock are connected the icon for the stock does not reflect the company icon in the asset and stock menu, nor in the stock exchange menu. It does show correctly in the swap modal.

There are console errors looking at the stock list

add trade stock button on the company list that selects the first stock, if it exists

PR #397, opened 2025-02-19
BLOCKCHAIN-586 Add update commission form

jindra12:


DorianSternVukotic:
You seem to have removed the stop validating button, or a way to switch back from validator to nominator

PR #398, opened 2025-02-19
Features/setup dark mode

jindra12:


src/index.scss
```diff
@@ -140,7 +133,14 @@ p {
 }
 
 .ant-card {
-  border-color: #ECEBF0!important;
+  @media screen and (prefers-color-scheme: light) {
+    border-color: #ECEBF0!important;
```
SzymonZiebaNeti, line 137: important is needed here?
if yes, why here is needed, but in dark mode, not?

src/index.scss
```diff
@@ -126,12 +114,17 @@ p {
...
 }
 .ant-tabs-tab-btn[aria-selected="true"] {
-  color: #243F5F !important;
+  color: var(--main-color)!important;
 }
 
 .description {
   font-weight: 600;
-  color: #ACBDC5;
+  @media screen and (prefers-color-scheme: light) {
```
SzymonZiebaNeti, line 122: If our web is default light mode, we need to checks for light and dark mode?
If we just used light colors by default and just use `prefers-color-scheme: dark`, it will be not enought?

SzymonZiebaNeti (commented):
Dark mode looking great - good job!

What about 
"add a darkmode switch in the menu"

Now darkmode is on, only when we have dark mode in system, what about this switcher?

SzymonZiebaNeti:
In figma, we have just one page designed for dark mode, but what you think about stay in this blue pallets colors everywhere?
Now modals, inputs, stock details, loader, etc. have black / dark grey background color. I think we should stay with ours blue colors, these greys, don't fit to other styles on page.
Maybe we should discuss it with more people, Dorian, Anton or others.

src/index.scss
jindra12: Simplified the styles, left required where needed, we now use variables for everything

src/index.scss
jindra12: Added a switch and removed all references from SCSS styles to scheme directly

PR #401, opened 2025-02-21
BLOCKCHAIN-596 Add LLD button info

jindra12:


DorianSternVukotic:
get LLD button on desktop on finances overview liquid LLD is too big

PR #402, opened 2025-02-22
Features/dark mode improvements

jindra12:


DorianSternVukotic:
Warning/info boxes are ugly and yellow background makes yellow links invisible

src/components/Profile/OnchainIdentityForm.js
```diff
@@ -126,12 +126,13 @@ function OnchainIdentityForm({
       )}
       {defaultValues.hasUserWarn && !isUserWarnAccepted && (
         <Paragraph>
-          <Alert
-            type="warning"
-            message={(
+          <Result
```
DorianSternVukotic, line 129: why call this result ?

jindra12: That is the name of the ANTD component https://ant.design/components/result/

PR #410, opened 2025-02-28
Blockchain 546 top taxpayers leaderboard and welfare wallofshame

SzymonZiebaNeti:


src/redux/reducers/offices.js
```diff
@@ -21,6 +21,7 @@ const initialState = {
   },
   backendAddressLLMBalance: null,
   pendingAdditionalMerits: [],
+  taxPayers: {},
```
jindra12, line 24: This should be unobtrusive

src/components/Offices/TaxPayers/index.js
```diff
@@ -0,0 +1,171 @@
...
+      ),
+    );
+  }, [dispatch, sortedPoolTotals, sortedUnpoolTotals, sortedTotalsByAddressPoolTotal]);
+
+  const handleSelectChange = (value) => {
+    setTimePeriodInMonth(value);
+  };
+
+  return (
+    <div>
```
jindra12, line 56: Maybe show a loader here for the data (<Spin />) before it loads + use unobtrusive loader

PR #417, opened 2025-04-14
Features/fix company registry request deletion

jindra12:


src/components/Companies/CompaniesForm/index.js
```diff
@@ -153,7 +155,9 @@ function getFieldComponent({
 }
 
 export const getDefaultValuesFromDataObject = (formObject, editMode = false) => {
-  const defaultValues = {};
+  const defaultValues = {
+    signedContract: Boolean(formObject.id),
```
DorianSternVukotic, line 159: Why is signedContract formobjectId?

jindra12: Because in order to create object with ID you need to submit signed form. Therefore, any form with ID is also signed

PR #454, opened 2025-08-22
BLOCKCHAIN-647 Hotfix validator returns

jindra12:


src/utils/staking.js
```diff
@@ -75,10 +75,10 @@ export const stakingInfoToProgress = (stakingInfo, progress) => {
 };
 
 const DEFAULT_PARAMS = {
-  falloff: 0.1,
+  falloff: 0.05,
   maxInflation: 0.1,
-  minInflation: 0.005,
-  idealStake: 0.75,
```
DorianSternVukotic, line 81: did you copy these numbers from polkadotjs? They are set for each chain, and LL chain has different settings.

PR #464, opened 2025-10-17
BLOCKCHAIN-648 Use broader resident api

jindra12:


DorianSternVukotic:
The point of this is to reduce theorhetical minimum onboarding time to one govt clerk confirmation instead of two.

With this, you can onboard onchain after paying. Previously you could only onboard after identity confirmation.

